PR #1, opened 2020-12-03
adding LICENSE.carbon

tpokorra:
see original Carbon License: https://github.com/carbon-design-system/carbon/blob/master/LICENSE
We have included the license in a similar way as Django does for Python: https://github.com/django/django/blob/master/LICENSE.python


saemideluxe (approved):

saemideluxe:
Thanks, this is an important one.

PR #12, opened 2021-02-24
support safety and do not fail on bandit

tpokorra:


saemideluxe (commented):
I see "safety" installed but never run. Is this intentionally? We would need to run it as ```safety check || true``` I think.

saemideluxe (changes requested):
I see "safety" installed but never run. Is this intentionally? We would need to run it as safety check || true I think.

saemideluxe (approved):

PR #23, opened 2021-04-29
Style email/phone-number/url input by adding an appropriate icon to the text field

wipascal:


bread/layout/components/email_input.py
```diff
@@ -0,0 +1,15 @@
+from htmlgenerator import DIV
+
+from bread.layout.components.icon import Icon
+from bread.layout.components.text_input import TextInput
+
+
+class EmailInput(TextInput):
+    def __init__(self, *args, **kwargs):
```
saemideluxe, line 8: Nitpicking: the variable ```icon``` could be avoided by directly passing the ```DIV``` as argument to append. This can be controversial (therefore my label "nitpicking"), but I think it reduces mental and perceptional load.

bread/layout/components/form.py
```diff
@@ -258,9 +261,10 @@ def _mapwidget(
```
saemideluxe, line 261: Good catch with the ```WIDGET_MAPPING```.
As already mentioned, this part of the code is one the more messy ones I think. But anyway, the PR changes are consistent and you were able work it out by yourself, so that is already something :)

bread/layout/components/phone_number_input.py
```diff
@@ -0,0 +1,15 @@
+from htmlgenerator import DIV
+
+from bread.layout.components.icon import Icon
+from bread.layout.components.text_input import TextInput
+
+
+class PhoneNumberInput(TextInput):
+    def __init__(self, *args, **kwargs):
```
saemideluxe, line 8: Same nitpick as above ;)

bread/static/design/carbon_design/scss/styles.scss
```diff
@@ -149,3 +149,21 @@ fieldset[readonly=true] :disabled {
 .highlighted {
     font-weight: 600;
 }
+
```
saemideluxe, line 152: Yes, I think totally justified, nice and generic implementation for the icon-feature.

saemideluxe (changes requested):
This looks all good and clean to me. I am not really experienced with formal code-reviews, but in general my system consists of three types of feedbacks: "Requires change", "nitpick" and 'generic comment".
"Nitpick" is just a nice additional kind of feedback I picked up somewhere, in order to tell: "That's how I would do it, but I will merge the PR if you prefer the way you submitted it."
The only thing required to do now would be to make sure back passes everying. In the overview of the PR you will see that the Github Action failed in the step for black formatting.

saemideluxe (approved):
All looks good

PR #24, opened 2021-04-29
fix issue with overlaping icons and PasswordInput (urgent. doesn't work at the moment)

wipascal:
- Fixed the issue with the overlapping icons.
- Shared code in common super-class `TextInputWithIcon`


bread/layout/components/text_input.py
```diff
@@ -39,10 +40,26 @@ def __init__(
...
         )
+
+        if icon is not None:
+            self.add_icon(icon)
+
         # for easier reference in the render method:
         self.label = self[0]
         self.input = self[1][0]
 
+    def add_icon(self, icon):
```
wipascal, line 51: @saemideluxe would you prefer to have this method inlined?

saemideluxe: If we inline the function I think it only improves the code if we put the the construct directly into the super().__init__ call.
The problem is that it is optional so we need to handle the branching somehow. Quickest and very likely cleanest solution would be to use htmlgenerator.If(icon, hg.DIV(......)) and the same for the class. I prefer it that way because you can see the whole potational tree of the component in one place. However, it is not always implemented that way (partly because I did not have the If-construct at the beginning and partly because sometimes it does not work well). So I recommend you take a look at the __init__method of a few other methods and decidec for yourself what is more readable and maintainable. My take would be: Even though a method ```add_icon``` is more intuitiv, it adds another "API" which the reader needs to "learn" and backtrack.

wipascal: Yes, this seems much better. An `add_icon()` method (whether inline or as an actual method) mutates the state after the object is constructed. If we can construct it the way we want it from the beginning, I prefer that. 

Btw. I am using pythons ugly ternary operator for the `text-input-with-icon` class. Is there a better way to do it? If not it's probably not too big of an issue.

saemideluxe: Agreed.
Regarding the ternary operator: First improvment would be to remove the ```is not None``` and use the implicit boolean casting of the ```if``` statement: ```text-input-with-icon" if icon else ""```.
The next viable option IMHO is to use ```htmlgenerator.If```. For all elements in the element tree the objects which inhert from ```htmlgenerator.HmlElement``` will render their attributes (normally passed in through named arguments of the __init__ method) as if they are part of the tree. This means we can do things like:

    hg.DIV("Hello World", _class=hg.BaseElement("class1", " class2", hg.If(True, "true", "false)))

Treating the branch as a part of the element tree is not always the nicest solution and tends to be a bit more verbose. But it is a bit more uniform with the rest of the tree construction and we can use lazy elements if necessary.

saemideluxe: Anyway, using ```text-input-with-icon" if icon else ""``` would also be appropiate I think.

wipascal: Alright, thanks.

wipascal:
This PR will conflict with that one: https://github.com/basxsoftwareassociation/bread/pull/26
I would prefer merging the other one first, and then resolving the conflicts on this branch.

saemideluxe (approved):
Okay, everything we discussed seems resolved. Thanks, good job with the two PRs.

PR #26, opened 2021-05-03
Make TextInput work without render() method

wipascal:
I am sorry that the git diff is so unreadable. This will maybe teach us the un-pythonic lesson to keep classes in separate files ;)

### DIFF
--- a/bread/layout/components/form.py
+++ b/bread/layout/components/form.py
@@ -330,7 +330,6 @@ def _mapwidget(
                 **data["attrs"],
                 **widgetattributes,
             }
-            return ret
 
         del kwargs["label"]
         del kwargs["required"]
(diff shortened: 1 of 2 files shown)


bread/layout/components/form.py
```diff
@@ -389,9 +388,14 @@ def extract_attrs(data):
         or WIDGET_MAPPING[type(field.field.widget)]
     )
     if isinstance(fieldtype, type) and issubclass(fieldtype, hg.BaseElement):
-        ret = fieldtype(
-            fieldname=field.name, widgetattributes=attrs, **elementattributes
-        )
```
saemideluxe, line 394: Adding these if statements all over the place in _mapwidget ist ugly but I think it is exactly the right choice until we have all form-components refactored to use the "new" style. And then we can simplify the logic inside _mapwidget :+1:

saemideluxe (approved):
Fantastically low lines added/removed ratio ;-)

PR #27, opened 2021-05-06
Refactor/form

wipascal:
Tried to improve the forms a bit

### DIFF
--- a/bread/layout/components/form.py
+++ b/bread/layout/components/form.py
@@ -309,17 +309,15 @@ def _mapwidget(
         return hg.DIV(
             *[
                 Checkbox(
-                    **{
-                        **elementattributes,
-                        "label": widget.data["label"],
-                        "widgetattributes": {
-                            "name": widget.data["name"],
-                            "value": widget.data["value"],
-                            "checked": widget.data["selected"],
-                            **widget.data["attrs"],
-                            **widgetattributes,
-                        },
-                    }
+                    **elementattributes,
+                    label=widget.data["label"],
+                    widgetattributes={
+                        "name": widget.data["name"],
+                        "value": widget.data["value"],
+                        "checked": widget.data["selected"],
+                        **widget.data["attrs"],
+                        **widgetattributes,
+                    },
                 )
                 for widget in field.subwidgets
             ],
--- a/bread/layout/components/form.py
+++ b/bread/layout/components/form.py
@@ -225,17 +225,15 @@ class InlineDeleteButton(FormChild, Button):
 
 
 class HiddenInput(FormChild, hg.INPUT):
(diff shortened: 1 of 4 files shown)


bread/layout/components/form.py
```diff
@@ -391,17 +368,14 @@ def extract_attrs(data):
...
+        ret = fieldtype(
+            fieldname=field.name,
+            widgetattributes=widgetattributes,
+            boundfield=field,
+            **elementattributes,
+        )
     else:
-        ret = fieldtype
-    ret.boundfield = field
+        raise Exception(f"Could not find a widget for field {field.name}")
```
saemideluxe, line 378: I am not sure raising an Exception is the best behaviour here. The point is that the framework becomes immediately unusable for any project which uses some 3rd-party widget or input field. Basically the same as PhoneNumberInput, where we had to add specialized behaviour. In general can go with two strategies here:

1. Raising an exception and and force the according developer to open an issue. We could then add the implementation for the widget somewhere in the code, would maybe need to make it a bit clearer which widgets are django-native and which belong to 3rd-party packages.
2. Having a half-dezent generalized solution which displays at least *something*. The widget might not work correctly, but at least it would show something and not break the application.

I agree with the idea that using an unsupported widget is an exception here. But for the sake of developer experience (people using this code) I tend to prefer 2.

bread/layout/components/form.py
```diff
@@ -412,3 +386,21 @@ def extract_attrs(data):
         )
 
     return ret
+
+
+def update_widgetattributes(field, only_initial, widgetattributes):
+    # TODO: This can be simplified, and improved
```
saemideluxe, line 392: Good idea throwing that messy code in a separate function :D

bread/layout/components/form.py
```diff
@@ -286,71 +281,50 @@ def _mapwidget(
...
                         "label": widget.data["label"],
-                        "widgetattributes": extract_attrs(widget.data),
+                        "widgetattributes": {
+                            "name": widget.data["name"],
+                            "value": widget.data["value"],
+                            "checked": widget.data["selected"],
+                            **widget.data["attrs"],
+                            **widgetattributes,
+                        },
                     }
```
saemideluxe, line 322: I suggest you remove the outer-most dict, so you can write a bit more compact:

    Checkbox(
        **elementattributes,
        label=widget.data["label"],
        widgetattributes={
            "name": widget.data["name"],
            "value": widget.data["value"],
            "checked": widget.data["selected"],
            **widget.data["attrs"],
            **widgetattributes,
        }
    )

saemideluxe (commented):
This is a great start for refactoring the forms-integration-code.

saemideluxe (approved):

PR #29, opened 2021-05-12
add: dataclass DataTableColumn instead of magic tuple

wipascal:
I personally think tuples are not the best thing for public APIs. So I created a class instead of the tuple.

### DIFF
--- a/bread/layout/components/datatable.py
+++ b/bread/layout/components/datatable.py
@@ -366,11 +366,11 @@ class DataTable(hg.BaseElement):
         page_urlparameter="page",
         itemsperpage_urlparameter="itemsperpage",
         settingspanel=None,
+        rowvariable="row",
         **kwargs,
     ):
         """TODO: Write Docs!!!!"""
         title = title or pretty_modelname(model, plural=True)
-        rowvariable = kwargs.get("rowvariable", "row")
 
         backquery = {"next": backurl} if backurl else {}
         if addurl is None:
--- a/bread/layout/components/datatable.py
+++ b/bread/layout/components/datatable.py
@@ -1,4 +1,3 @@
-from dataclasses import dataclass
 from typing import Any, List, NamedTuple, Optional, Union
 
 import htmlgenerator as hg
--- a/bread/layout/components/datatable.py
+++ b/bread/layout/components/datatable.py
@@ -60,7 +60,7 @@ def sortinglink_for_column(orderingurlparameter, columnname):
     return aslink_attributes(hg.F(extractsortinglink))
 
 
-@dataclass
+@dataclass(frozen=True)
 class DataTableColumn:
     header: Any
     cell: Any
--- a/bread/layout/components/datatable.py
+++ b/bread/layout/components/datatable.py
@@ -1,5 +1,5 @@
 from dataclasses import dataclass
-from typing import Any, List, Union
+from typing import Any, List, NamedTuple, Union
 
 import htmlgenerator as hg
 from django.db import models
@@ -60,8 +60,7 @@ def sortinglink_for_column(orderingurlparameter, columnname):
     return aslink_attributes(hg.F(extractsortinglink))
 
 
-@dataclass(frozen=True)
-class DataTableColumn:
+class DataTableColumn(NamedTuple):
     header: Any
     cell: Any
     sortingname: str = None
--- a/bread/layout/components/datatable.py
+++ b/bread/layout/components/datatable.py
@@ -66,15 +66,16 @@ class DataTableColumn(NamedTuple):
     enable_row_click: bool = True
 
     @staticmethod
-    def from_modelfield(col, model, prevent_automatic_sortingnames, rowvariable):
-        col = DataTableColumn(
+    def from_modelfield(
+        col, model, prevent_automatic_sortingnames, rowvariable
+    ) -> "DataTableColumn":
+        return DataTableColumn(
             fieldlabel(model, col),
             hg.C(f"{rowvariable}.{col}"),
             sortingname_for_column(model, col)
             if not prevent_automatic_sortingnames
             else None,
         )
-        return col
 
 
 class DataTable(hg.BaseElement):
--- a/bread/layout/components/datatable.py
+++ b/bread/layout/components/datatable.py
@@ -66,6 +66,17 @@ class DataTableColumn(NamedTuple):
     sortingname: Optional[str] = None
     enable_row_click: bool = True
 
+    @staticmethod
+    def from_modelfield(col, model, prevent_automatic_sortingnames, rowvariable):
+        col = DataTableColumn(
+            fieldlabel(model, col),
+            hg.C(f"{rowvariable}.{col}"),
+            sortingname_for_column(model, col)
+            if not prevent_automatic_sortingnames
+            else None,
+        )
+        return col
+
 
 class DataTable(hg.BaseElement):
     SPACINGS = ["default", "compact", "short", "tall"]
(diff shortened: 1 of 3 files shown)


saemideluxe:
> I personally think tuples are not the best thing for public APIs. So I created a class instead of the tuple.

Agreed!

bread/layout/components/datatable.py
```diff
@@ -57,19 +60,27 @@ def extractsortinglink(context, element):
...
 
+@dataclass
+class DataTableColumn:
+    header: Any
+    cell: Any
+    sortingname: str = None
+    enable_row_click: bool = True
+
+
 class DataTable(hg.BaseElement):
```
saemideluxe, line 71: Since we deal with immutable data here NamedTuple might be preferable here. What do you think?

bread/menu.py
```diff
@@ -100,6 +101,18 @@ def from_objectaction(object, actionname, label, icon=None, *args, **kwargs):
...
+        from . import layout
+
+        super().__init__(
+            js=hg.F(
+                lambda c, e: f'window.location = \'{layout.objectaction(c["row"], "delete")}?next=\' + window.location.pathname + window.location.search',
+            ),
+            icon="trash-can",
+        )
+
+
```
saemideluxe, line 115: Can you explain the purpose of this class? I do not see it used in this PR?
I wonder if this is the right place to add this shortcut. I definitely see the need for this kind of shortcut to make the API simpler. But the problem with these kind of "object-action" is that they depend at least on the context variable of the according object ("row" in this case) and the availability of an object-action (or URL in the end), here "delete". I think the ```hg.F(....)``` should be subclassed in a generic way (that means it would be a ```htmlgenerator.Lazy``` object which can handle object-actions).
In any case I would keep the ```menu``` module free of dependencies to the ```layout``` module for now. The menu- and action-system anyway needs to be redesigned. One point is that the menu itself could easier be implemented by having the corresponding carbon design menu and then just adding that element to the template/render context. Another point is that Action and Link need possibly to be implemented inside the ```layout``` module. They both represent the abstract idea of an action a user can do in the UI. Decoupling that idea completely from the UI-code (the ```layout``` module) is likely not a very good idea.
But in any case, this would be a bit a larger refactoring, I just braindumped a bit here so you also get aware of the issue.

saemideluxe:
Good job with all the small improvments inside the datatables code.

bread/layout/components/datatable.py
saemideluxe: Also, see my comment on https://github.com/basxsoftwareassociation/basxconnect/pull/68#pullrequestreview-658555347 regarding an additional helper-method to prevent the use of pure strings. It might simplify some of the data-tables code. I copy-paste it here:

> Looks good to me, much more clear with the DataTableColumn. If we already use DataTableColumn, should we maybe enforce it and prohibit the use of plain strings? A static-method-shortcut like DataTableColumn.from_modelfield(fieldname) would be interesting I think.

And we could also check the columns argument in ```DataTable.__init__``` to be of the correct type (and then people say "duuhhh, that is what statically typed languages are for" :D and I would agree). There is already some type checking happening in the DataTables code, to catch misconfigured datatables early.

bread/layout/components/datatable.py
```diff
@@ -57,19 +60,27 @@ def extractsortinglink(context, element):
     return aslink_attributes(hg.F(extractsortinglink))
 
 
+@dataclass
+class DataTableColumn:
+    header: Any
+    cell: Any
+    sortingname: str = None
```
saemideluxe, line 67: For sortingname we could use ```typing.Optional[str]```

bread/layout/components/datatable.py
wipascal: I agree that no one should mutate an instance of this class ;) . I also thought of the named tuple, but when I saw the syntax it looked a lot like the advanced reflection-like features of python that I prefer not using. (Declaring the name of a field using a string...). Maybe I should add the `frozen=True` annotation. But then, it feels to go against the Python mindset to _enforce_ immutability. I think in Python people need to act responsibly, otherwise we are heading to disaster anyway... 

I like the idea of a static `DataTableColumn.from_modelfield(fieldname, model)` method!

To be honest, I am against these type checks in Python because type hints together with static analysis detect this kind of errors without adding boilerplate code for type checking. If we _have_ to enforce types, I'd much rather use a decorator like this one: https://stackoverflow.com/a/32685519 . What I love about Python is that there is almost no boilerplate, which potentially makes code very readable, so I'd rather not give that up.

bread/menu.py
wipascal: I don't even understand this code fully :D
I just needed a "delete"-Button for each row and found something similar somewhere else in the code so I thought I am going to share the code in order to make explicit that I am doing the exact same thing (without understanding it).
I guess what I will do is to remove the Delete class and duplicate the F function. Are you ok with that?

bread/layout/components/datatable.py
saemideluxe: The immutability thing is not super important I think, and yes, python programmers are not supposed to behave too stupid :D. In any case we could use ```typing.NamedTuple``` this is not using strings (at least not directly, when running the code it is anyways because the types are in all cases constructed at runtime) and we could also use type hints with that.

Yes, the check of string vs DataTableColumn is not necessairily important, so I am fine without that anyways.

However, I should add: The problem with the "only type hints and static analysis" approach (or any type system for that matter) is that it is never expressiv enough to do all the checks for you. I think it is comparable to statically typed languages to some degree: Before "Nullable" became a thing in C# and Java we wrote a lot of "dynamic type checking" with ```if(arg1 == null) raise IllegalParameterException("Arg1 cannot be null")``` and similar.
Now, another problem with type hints in Python is that type hints are (and will always be, as far as I understand the propsal) only comment to the interpreter and do not influence the behaviour of the program in any way. This means type hints only work for us in our projects if you have mypy or similar integrated in the development workflow. Any other programer using this code (and this is crucial for framework and library code) will need to debug things if he doesn't run a type-checker first.
Therefore, I suggest that dynamic integrity checks are a very helpfull thing and we should keep other users of our APIs in mind.
Maybe another approach to look at it, more generally, would be: It is often helpfull to ensure certain invariants are true in order to make the rest of the code simpler because we can always rely on these invariants being true.

wipascal: Oh great, I didn't know about `typing.NamedTuple`! I like the syntax of that one :)
I agree that static analysis doesn't catch everything and that for stuff it doesn't catch, dynamic checks are very useful.
I just assumed (and hoped) Python is moving to a direction where most people will notice if their usage of a function and the type hints are not consistent. If this is not the case (and you know that better than me), I am happy to check the types manually.

saemideluxe: Right! So, as I just mentioned in the ```htmlgenerator``` PR: Type hinting is still very unadapted. I think we can profit from from more type hints. But, I would still keep it a bit on the "exploring" side, also for this project. In Python it is often good to follow the mainstream a bit, be it for tooling, readability or just compatability (the black-formatter was an interesting example of seeing that kind of "adoption" process throughout the wider Python community).

The ```htmlgenerator``` project is a good place to explore because it is a very standalone and small code base.
I think if we want to do more with type hinting inside bread we should integrate it into the GitHub CI and maybe start with smaller packages or modules which are more standalone. PM me if you want add more of it in existing code, I can give you a hint where to start.

wipascal:
I made the mistake to merge a PR in basxconnect which depends on that one... So it would be good to merge it soon :)

saemideluxe:
> I made the mistake to merge a PR in basxconnect which depends on that one... So it would be good to merge it soon :)

Np, if you blocks your workflow feel free to force-merge next time.

saemideluxe (approved):

PR #31, opened 2021-05-19
Feature/UI restructuring

wipascal:


saemideluxe (changes requested):
Could we add a short comment why ```.theme-white``` is needed? If it is only required for the person-view we could argue that the style should only be added in the basxconnct repo. But I understand from the meeting with Frederik that this kind of layout is used when a lot information is on display on a single page. If that is the use-case of ```.theme-white``` I would mention it somehow.

bread/layout/components/tabs.py
```diff
@@ -50,9 +60,10 @@ def __init__(self, *tabs, container=False, **attributes):
             ),
             self.tablabels,
             data_tabs=True,
-            _class="bx--tabs" + (" bx--tabs--container" if container else ""),
+            _class=" bx--tabs" + ("bx--tabs--container" if container else ""),
```
wipascal, line 63:
```suggestion
            _class="bx--tabs" + (" bx--tabs--container" if container else ""),
```

saemideluxe (approved):

PR #32, opened 2021-05-21
add namedtuple Tab and comment which I commited on the wrong branch

wipascal:
sorry, I did add the comment which you suggested to add (about .theme-white), but I commited it on the wrong branch 😄 

I am only using this version of the namedtuple because I didn't want to add more type hints before adding mypy to the CI. I find `typing.namedtuple` much more readable.

the Tab namedtuple is used here: https://github.com/basxsoftwareassociation/basxconnect/pull/75

### DIFF
--- a/bread/layout/components/tabs.py
+++ b/bread/layout/components/tabs.py
@@ -36,6 +36,9 @@ class TabPanel(hg.DIV):
         )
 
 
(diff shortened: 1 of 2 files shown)


saemideluxe (approved):
Ok, thanks for the comment.

PR #33, opened 2021-05-21
Make full width the default grid mode

wipascal:


saemideluxe (approved):
"full-width" is the most used one I guess?

wipascal:
> "full-width" is the most used one I guess?

yes, that's how I interpret the carbon documentation 😄

PR #34, opened 2021-05-22
Features/formsets in datatables

saemideluxe:
The code introduced a bit of dependency between Datatable and FormsetField but I think it is good enough. Better solutions would require too much refactoring.

### DIFF
--- a/bread/layout/components/datatable.py
+++ b/bread/layout/components/datatable.py
@@ -98,7 +98,7 @@ class DataTable(hg.BaseElement):
         sortingname: value for the URL parameter 'orderingurlparameter', None if sorting is not allowed
         spacing: one of "default", "compact", "short", "tall"
         zebra: alternate row colors
-        is_inlineformset: indicates that row_iterator is the name of model form field and not hg.Iterator but .form.FieldsetField will be used for iteration
+        is_inlineformset: indicates that row_iterator is the name of model form field and not hg.Iterator but .form.FormsetField will be used for iteration
         """
         if spacing not in DataTable.SPACINGS:
             raise ValueError(
--- a/bread/layout/components/datatable.py
+++ b/bread/layout/components/datatable.py
@@ -89,6 +89,7 @@ class DataTable(hg.BaseElement):
         spacing="default",
         orderingurlparameter="ordering",
         zebra=False,
+        is_inlineformset=False,
     ):
         """columns: list of DataTableColumn(header, cell, sortingname, enable_row_click)
         row_iterator: python iterator of htmlgenerator.Lazy object which returns an iterator
@@ -97,6 +98,7 @@ class DataTable(hg.BaseElement):
         sortingname: value for the URL parameter 'orderingurlparameter', None if sorting is not allowed
         spacing: one of "default", "compact", "short", "tall"
         zebra: alternate row colors
+        is_inlineformset: indicates that row_iterator is the name of model form field and not hg.Iterator but .form.FieldsetField will be used for iteration
         """
         if spacing not in DataTable.SPACINGS:
             raise ValueError(
@@ -133,16 +135,21 @@ class DataTable(hg.BaseElement):
                 hg.TH(headcontent, **getattr(col.header, "td_attributes", {}))
             )
 
-        self.iterator = hg.Iterator(
-            row_iterator,
-            rowvariable,
-            hg.TR(
-                *[
-                    hg.TD(col.cell, **getattr(col.cell, "td_attributes", {}))
-                    for col in columns
-                ]
-            ),
+        row_template = hg.TR(
+            *[
+                hg.TD(col.cell, **getattr(col.cell, "td_attributes", {}))
+                for col in columns
+            ]
         )
+
+        # combining Formsets with datatables needs to handle iteratons differently
+        if is_inlineformset:
+            # need local import here to prevent circular import
+            from .form import FormsetField
+
+            self.iterator = FormsetField(row_iterator, row_template)
+        else:
+            self.iterator = hg.Iterator(row_iterator, rowvariable, row_template)
         super().__init__(
             hg.TABLE(
                 hg.THEAD(self.head),
--- a/bread/layout/components/datatable.py
+++ b/bread/layout/components/datatable.py
@@ -143,17 +143,22 @@ class DataTable(hg.BaseElement):
         )
 
         # combining Formsets with datatables needs to handle iteratons differently
+        # the formsetfield needs to control the container in order to attach new entires
         if is_inlineformset:
             # need local import here to prevent circular import
             from .form import FormsetField
 
-            self.iterator = FormsetField(row_iterator, row_template)
+            self.iterator = FormsetField(
+                row_iterator, row_template, containertag=hg.TBODY
+            )
+            body = self.iterator
         else:
             self.iterator = hg.Iterator(row_iterator, rowvariable, row_template)
+            body = hg.TBODY(self.iterator)
         super().__init__(
             hg.TABLE(
                 hg.THEAD(self.head),
-                hg.TBODY(self.iterator),
+                body,
                 _class=" ".join(classes),
             )
         )
--- a/bread/layout/components/datatable.py
+++ b/bread/layout/components/datatable.py
@@ -1,4 +1,4 @@
-from typing import Any, List, NamedTuple, Optional, Union
+from typing import Any, Iterable, List, NamedTuple, Optional, Union
 
 import htmlgenerator as hg
 from django.db import models
@@ -14,49 +14,6 @@ from .icon import Icon
 from .overflow_menu import OverflowMenu
 from .pagination import Pagination
 from .search import Search
-
-
-def sortingclass_for_column(orderingurlparameter, columnname):
-    def extracturlparameter(context, element):
-        value = context["request"].GET.get(orderingurlparameter, "")
-        if not value:
-            return ""
-        if value == columnname:
-            return "bx--table-sort--active"
-        if value == "-" + columnname:
-            return "bx--table-sort--active bx--table-sort--ascending"
-        return ""
-
-    return hg.F(extracturlparameter)
-
-
-def sortingname_for_column(model, column):
-    components = []
-    for field in resolve_modellookup(model, column):
-        if hasattr(field, "sorting_name"):
-            components.append(field.sorting_name)
-        elif isinstance(field, models.Field):
-            components.append(field.name)
-        else:
-            return None
-    return "__".join(components)
-
-
-def sortinglink_for_column(orderingurlparameter, columnname):
-    ORDERING_VALUES = {
-        None: columnname,
-        columnname: "-" + columnname,
-        "-" + columnname: None,
-    }
-
-    def extractsortinglink(context, element):
-        currentordering = context["request"].GET.get(orderingurlparameter, None)
-        nextordering = ORDERING_VALUES.get(currentordering, columnname)
-        return link_with_urlparameters(
-            context["request"], **{orderingurlparameter: nextordering}
-        )
-
-    return aslink_attributes(hg.F(extractsortinglink))
 
 
 class DataTableColumn(NamedTuple):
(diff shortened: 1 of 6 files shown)


bread/static/js/main.js
```diff
@@ -6,7 +6,10 @@ document.addEventListener(
 );
 
 function updateMultiselect(e) {
-    $('.bx--list-box__selection', e).firstChild.textContent = $$('fieldset input[type=checkbox][checked]', e).length;
+    let elem = $('.bx--list-box__selection', e);
+    if (elem) {
+        elem.firstChild.textContent = $$('fieldset input[type=checkbox][checked]', e).length;
+    }
 }
```
saemideluxe, line 13: This is an unrelated fix to prevent a javascript exception in the console showing up

bread/layout/components/datatable.py
```diff
@@ -89,6 +89,7 @@ def __init__(
         spacing="default",
         orderingurlparameter="ordering",
         zebra=False,
+        is_inlineformset=False,
     ):
         """columns: list of DataTableColumn(header, cell, sortingname, enable_row_click)
         row_iterator: python iterator of htmlgenerator.Lazy object which returns an iterator
```
wipascal, line 95:
```suggestion
        row_iterator: python iterator or htmlgenerator.Lazy object which returns an iterator or a string which we need to forward to a `FormsetField` constructor in case `is_inlineformset` is true
```
I am not a fan of parameters with multiple meanings anyway. Can we not just have two static factory-methods (which both call the constructor, which contains the shared logic)? One with the FormsetField string as parameter and one with the Iterator/Lazy[Iterator]?

bread/layout/components/datatable.py
```diff
@@ -133,20 +135,30 @@ def __init__(
...
-            ),
+        row_template = hg.TR(
+            *[
+                hg.TD(col.cell, **getattr(col.cell, "td_attributes", {}))
+                for col in columns
+            ]
         )
+
+        # combining Formsets with Datatables needs to handle iterations differently,
+        # the formsetfield needs to control the container in order to attach new entries/rows
```
wipascal, line 146: that's a very useful comment, thanks!
(hopefully not needed anymore if we can have a cleaner design as I suggested in other comments)

bread/layout/components/form.py
```diff
@@ -162,23 +167,72 @@ def render(self, context):
...
 
+    @staticmethod
+    def as_datatable(
+        fieldname,
+        inlinefieldnames,
+        formname="form",
+        title=None,
+        **formsetfactory_kwargs,
+    ):
+        """
```
wipascal, line 197: that's useful documentation, thanks

bread/layout/components/form.py
```diff
@@ -162,23 +167,72 @@ def render(self, context):
...
+        :param str fieldname: The fieldname which should be used for an formset, in general a n-to-many field
```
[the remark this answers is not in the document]
saemideluxe: What I meant was n-to-many with n &#8712; {one, many}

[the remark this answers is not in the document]
saemideluxe: Changed, but in the wrong branch (refactoring/bulk-actions-url-parameter-handling): a54952f

bread/layout/components/form.py
```diff
@@ -162,23 +167,72 @@ def render(self, context):
...
             ),
-        )
-        yield from skeleton.render(context)
+        ).render(context)
 
     def __repr__(self):
         return f"Formset({self.fieldname}, {self.formsetfactory_kwargs})"
 
+    @staticmethod
+    def as_datatable(
```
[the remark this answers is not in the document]
saemideluxe: I had it there first but it felt wrong that DataTables code needs to know that much details about form handling. The form.py module is on a higher abstraction level than datatable (it should maybe not even be inside the "components" module). Therefore I put the as_datatable into the forms.

bread/layout/components/form.py
```diff
@@ -162,23 +167,72 @@ def render(self, context):
...
+        title=None,
+        **formsetfactory_kwargs,
+    ):
+        """
+        :param str fieldname: The fieldname which should be used for an formset, in general a n-to-many field
+        :param list inlinefieldnames: A list of strings or objects. Strings are converted to DataTableColumn, objects are passed on as they are
+        :param str formname: Name of the surounding django-form object in the context
+        :param str title: Datatable title, automatically generated from form if None
+        :return: A datatable with inline-editing capabilities
+        :rtype: hg.HTMLElement
```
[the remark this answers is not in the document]
saemideluxe: Because if we wrap with toolbar it is hg.DIV and not DataTable. I guess we could use hg.HTMLTag (which I had first) but it seems to be less clear what you actually get...

bread/layout/components/datatable.py
```diff
@@ -133,20 +135,30 @@ def __init__(
...
+        if is_inlineformset:
+            # need local import here to prevent circular import
+            from .form import FormsetField
```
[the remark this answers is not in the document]
saemideluxe: Okay, continuing from above (because I suggested the opposite from your argument here :D).

If you use "isolated" and "dependent" instead of low-level and high-level the situation looks like this:

- form.py is highly dependant on Django implementation details (of the HTML forms). It is much more a logical component than a visual one.

- The other components should are very isolated and should only care about visual representation, not about django internals. Now, admittedly the datatable module has an ugly dependency to the models-system of Django for the from_model etc., but I would prefer to factor that out than to include even more logical dependencies in datatable.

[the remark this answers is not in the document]
wipascal: Ok, I get it :)
I am fine with form depending on DataTable and not the other way around.
But then, datatable should have a constructor or factory method which accepts a FormsetField. It should not create the FormsetField itself...

bread/layout/components/datatable.py
```diff
@@ -97,6 +98,7 @@ def __init__(
         sortingname: value for the URL parameter 'orderingurlparameter', None if sorting is not allowed
         spacing: one of "default", "compact", "short", "tall"
         zebra: alternate row colors
+        is_inlineformset: indicates that row_iterator is the name of model form field and not hg.Iterator but .form.FieldsetField will be used for iteration
```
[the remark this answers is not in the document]
saemideluxe: Right, fixed that.

wipascal (approved):
I don't want to block you from merging at all. Just go ahead as soon as you are confident with the changes.
As for me, I strongly prefer having a constructor which takes the `self.iterator`-thingy (or should we call it `table_body`?) as argument. This constructor can then document what this thing actually is (and there is no need even to mention `FormsetField`, as the `DataTable` does not know anything about `FormsetField`s).

bread/layout/components/datatable.py
```diff
@@ -16,49 +16,6 @@
...
-    def extractsortinglink(context, element):
-        currentordering = context["request"].GET.get(orderingurlparameter, None)
-        nextordering = ORDERING_VALUES.get(currentordering, columnname)
-        return link_with_urlparameters(
-            context["request"], **{orderingurlparameter: nextordering}
-        )
-
-    return aslink_attributes(hg.F(extractsortinglink))
-
-
```
saemideluxe, line 61: I just moved these to the bottom to make parsing the file easier for humans.

bread/layout/components/datatable.py
```diff
@@ -77,79 +34,63 @@ def from_modelfield(
...
-                    hg.TD(col.cell, **getattr(col.cell, "td_attributes", {}))
-                    for col in columns
-                ]
-            ),
-        )
-        super().__init__(
-            hg.TABLE(
-                hg.THEAD(self.head),
-                hg.TBODY(self.iterator),
-                _class=" ".join(classes),
```
saemideluxe, line 150: These lines have mostly been put in a few staticmethods at the bottom.

bread/views/read.py
```diff
@@ -49,9 +49,11 @@ def layoutasreadonly(layout):
...
             isinstance(
                 element,
-                (_layout.form.InlineDeleteButton, _layout.form.FormsetAddButton),
+                hg.BUTTON,
             )
             or getattr(element, "attributes", {}).get("type") == "submit"
         )
     )
+    for form in layout.filter(lambda element, ancestors: isinstance(element, hg.FORM)):
+        form.attributes["onsubmit"] = "return false;"
```
saemideluxe, line 58: This is a small improvment to prevent changes beeing submitted "accidentally" on the read-view. Browser seem to submit a form when any button (or the first button) is clicked inside the form if no submit button has been defined. This is a bit problematic because e.g. the carbon tags use a button element for representation.

saemideluxe:
Ok, DataTable and FormsetField APIs feel much better now. It should now also be possible to compose easier in other scenarios.

bread/layout/components/datatable.py
```diff
@@ -77,79 +34,63 @@ def from_modelfield(
...
-        spacing: one of "default", "compact", "short", "tall"
-        zebra: alternate row colors
+        """A carbon DataTable element
+
+        :param columns: Column definitions
+        :param row_iterator: Iterator which yields row objects. If this is a hg.Iterator instance it will be used for the table body, otherwise a default Iterator will be used to render the column cells.
+        :param rowvariable: Name of the current object passed to childrens context
+        :param orderingurlparameter: The name of the GET query parameter which is used to set the table ordering
+        :param spacing: One of "default", "compact", "short", "tall", according to the carbon styles
+        :parma zebra: If True alternate row colors
```
wipascal, line 79:
```suggestion
        :param zebra: If True alternate row colors
```

bread/layout/components/datatable.py
```diff
@@ -77,79 +34,63 @@ def from_modelfield(
...
+        :parma kwargs: HTML element attributes
```
wipascal, line 80:
```suggestion
        :param kwargs: HTML element attributes
```

bread/layout/components/form.py
```diff
@@ -115,89 +120,187 @@ def render(self, context):
...
 
-    def render(self, context):
-        formset = self.form[self.fieldname].formset
-        self.attributes["id"] = f"add_{formset.prefix}_button"
-        self.attributes[
-            "onclick"
-        ] = f"formset_add('{ formset.prefix }', '#formset_{ formset.prefix }_container');"
-        return super().render(context)
+    @staticmethod
+    def as_plain(*args, add_label=_("Add"), **kwargs):
```
wipascal, line 228: Maybe a name like `as_textinput_group` would be more descriptive.

wipascal:
This is great! :)
Thank you for taking the time to refactor!

bread/layout/components/datatable.py
```diff
@@ -77,79 +34,63 @@ def from_modelfield(
...
 
     def __init__(
         self,
         columns: List[DataTableColumn],
-        row_iterator,
-        rowvariable="row",
-        spacing="default",
-        orderingurlparameter="ordering",
-        zebra=False,
+        row_iterator: Union[hg.Lazy, Iterable, hg.Iterator],
```
wipascal, line 65: Seing the options of `Iterable` and `hg.Iterator` makes it very clear to me what is going on. I love that.
What still feels a bit like a black box to me are these `hg.Lazy` dudes...

bread/layout/components/datatable.py
```diff
@@ -77,79 +34,63 @@ def from_modelfield(
...
-        row_iterator: python iterator of htmlgenerator.Lazy object which returns an iterator
-        rowvariable: name of the current object passed to childrens context
-        if the header_expression/row_expression has an attribute td_attributes it will be used as attributes for the TH/TD elements (necessary because sometimes the content requires additional classes on the parent element)
-        sortingname: value for the URL parameter 'orderingurlparameter', None if sorting is not allowed
-        spacing: one of "default", "compact", "short", "tall"
-        zebra: alternate row colors
+        """A carbon DataTable element
+
+        :param columns: Column definitions
+        :param row_iterator: Iterator which yields row objects. If this is a hg.Iterator instance it will be used for the table body, otherwise a default Iterator will be used to render the column cells.
```
wipascal, line 75: I am not sure if it would help to mention also the `hg.Lazy` option here or if I just need to understand it better and then it will be clear ;)

wipascal: Now I read the original comment saying `...or htmlgenerator.Lazy object which returns an iterator`. Maybe we can just put that one back in and then it's clear.

bread/layout/components/datatable.py
wipascal: As the "java guy" here I would not be against seeing someting like `hg.Lazy[Iterable]` anyway 😄

bread/layout/components/datatable.py
saemideluxe: Okay, will add it again. Maybe we need a better way to mark support for ```Lazy``` values.

Lazy values are just values which are only evaluated very late, meaning at the time when the whole tree is serialized to an HTML string. A value which is lazy-generated cannot be modified or "queried" in the tree, that is why we want to keep them as rather small units. In most cases the lazy values mark that a value is only "extractable" or accessible while having an http-request from the framework available. In general these "values" are accessible through the context, which we passe when we call "render" on the html element tree, where we pass the django-context which contains at least the request object and often more (like the current database object).

bread/layout/components/datatable.py
saemideluxe: Would ```hg.Lazy[Iterable]``` be comprehensible for a type checker? I think we would need to define it as container-type or something...

bread/layout/components/datatable.py
saemideluxe: Must have been dreaming about cheese :cheese: :D

bread/layout/components/form.py
saemideluxe: Why textinput?

PR #36, opened 2021-05-26
white background for Stammdaten UI, border for navbar, ...

wipascal:
Before that I achieved the full-width white background using negative margins. This is now an ugly CSS class, but at least it comes from the official tutorial :D


saemideluxe (approved):
I wouldn't say I am competent enough in CSS/SCSS to judge carbon-CSS related changes, so I will just approve this ;)
But anyway  good decision to keep all custom changes in this file, upgrading to v11 will be less painless this way.

PR #41, opened 2021-06-03
fix: kwargs are not of type dict

wipascal:
pycharm tells me dict is not the right annotation...
we would need to put the type of the individual parameters. I think we can just leave it empty, as I beliebe it is not standard to use typing.Any in docstrings. (Once we use typing.* we can just as well use actual type hints.)


saemideluxe (approved):
Ok

PR #44, opened 2021-06-08
field was not style-able via elementattributes, "attributes" argument…

saemideluxe:
… should be usable for HTMLTag.

I am not sure if we need to re-check all the input-element ```__init__``` methods.

### DIFF
--- a/bread/layout/components/text_input.py
+++ b/bread/layout/components/text_input.py
@@ -19,6 +19,13 @@ class TextInput(hg.DIV):
         icon=None,
         **attributes,
     ):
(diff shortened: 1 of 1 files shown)


bread/layout/components/text_input.py
```diff
@@ -19,6 +19,11 @@ def __init__(
         icon=None,
         **attributes,
     ):
+        attributes.pop("boundfield")
```
wipascal, line 22: I would appreciate a comment on why these are popped.

wipascal (changes requested):
coo

wipascal (approved):

PR #45, opened 2021-06-10
add SearchSelect input widget

wipascal:


wipascal:
now ready for review @saemideluxe

bread/layout/components/form.py
```diff
@@ -98,6 +98,8 @@ def __init__(
         widgetattributes=None,
         formname="form",
     ):
+        if fieldtype is not None and not isinstance(fieldtype, type):
+            raise ValueError(f"argument 'fieldtype' has is not a type")
```
saemideluxe, line 102: ```f"argument 'fieldtype' is not a type")```

bread/layout/components/form.py
```diff
@@ -478,6 +464,30 @@ def _mapwidget(
...
+        forms.SelectMultiple: MultiSelect,  # TODO HIGH
+        forms.RadioSelect: TextInput,  # TODO HIGH
+        forms.FileInput: FileUploader,
+        forms.ClearableFileInput: FileUploader,  # TODO HIGH
+        forms.MultipleHiddenInput: TextInput,  # TODO
+        forms.SplitDateTimeWidget: TextInput,  # TODO
+        forms.SplitHiddenDateTimeWidget: TextInput,  # TODO
+        forms.SelectDateWidget: TextInput,  # TODO
+    }
+
```
saemideluxe, line 490: Every PR improves _mapwidget a little bit more, until, one day, it is gooone :D

bread/layout/components/search_select.py
```diff
@@ -0,0 +1,109 @@
...
+
+from .icon import Icon
+from .loading import Loading
+from .tag import Tag
+
+
+class SearchSelect(hg.BaseElement):
+    def __init__(
+        self,
+        search_view,
```
saemideluxe, line 14: I think this should be a URL and not a viewname. Using reverse inside a layout component is likely suboptimal because we can only use django-urls without parameters. But we might want to use an external URL or a django view with parameters in the path. These decisions should in general not be done in the components if we can avoid. The caller/user should make this decision.

bread/layout/components/search_select.py
```diff
@@ -0,0 +1,109 @@
...
+            getattr(
+                elementattributes["boundfield"].form.instance,
+                elementattributes["fieldname"],
+            )
+            if current_selection_id
+            else None
+        )
+        del elementattributes["boundfield"]
+        elementattributes["_class"] = (
+            elementattributes.get("_class", "") + f" bx--search bx--search--{size}"
```
saemideluxe, line 33: The class ```bx--search--{size}``` should be on the input element, not on the outer element. I saw this was originally my fault, done in ```search.py```.

bread/layout/components/search_select.py
```diff
@@ -0,0 +1,109 @@
...
+        current_selection_id = widgetattributes["value"][0]
+        # This works inside a formset. Might need to be changed for other usages.
+        current_selection = (
+            getattr(
+                elementattributes["boundfield"].form.instance,
+                elementattributes["fieldname"],
+            )
+            if current_selection_id
+            else None
+        )
```
saemideluxe, line 30: I think this code would be better of with Lazy elements, refering to my big "form elements refactoring" which never happens ;).
When we depend on implementation details from boundfield we can not use the component outside of a Django form. But yeah, it is a mess currently and also not a big issue for now. But it would help in general to decouple the components a bit in the future.

bread/layout/components/search_select.py
```diff
@@ -0,0 +1,109 @@
...
+        }
+
+        resultcontainerid = f"search-result-{hg.html_id((self, search_view))}"
+
+        url = bread.utils.reverse(
+            search_view,
+            query={
+                "target-id-to-store-selected": widgetattributes["id"],
+            },
+        )
```
saemideluxe, line 53: As mentioned above: Possible to move into basxconnect?

bread/static/js/main.js
```diff
@@ -128,6 +128,13 @@ function formset_add(form_prefix, list_container) {
...
     update_add_button(form_prefix);
     updateMultiselect(container_elem);
+
+    htmx.process(container_elem);
+}
+
+function set_value(selector, value) {
+    let element = $(selector)
+    element.value = value
 }
```
saemideluxe, line 138: Not sure why this is needed, see comment in basxconnect PR.

saemideluxe (changes requested):
Looks good, mostly the URL/view issue which should be resolved.

bread/layout/components/search_select.py
wipascal: are you sure? see screenshot from the official examples
grafik

bread/layout/components/search_select.py
wipascal: The problem is that this ID needs to be unique for each `SearchSelect` widget that is created. But the widgets themselves are not created in basxconnect, but in `_mapwidget`...

bread/layout/components/form.py
wipascal: 😄

bread/static/js/main.js
wipascal: you're right, it is better to inline that one

bread/layout/components/search_select.py
saemideluxe: Sorry, my confusion. I misread the SCSS definition.

bread/layout/components/search_select.py
saemideluxe: Okay, I would like us to work a bit on this one.

Apart from the viewname vs URL. issue we also have the problem that the name of the query parameter needs to be defined/hardcoded in two different places, once in ```layout/components/search_select.py``` and once in ```basxconnect/core/views/person/person_search_views.py```.

Try to decouple the search component from the search view a bit more would improve things I suggest. How about something like this:

- The search view can return whatever it wants to be. We just need to make sure that the pk-values and the labels are extractable via CSS selectors.
- The search component takes as ```__init__``` parameter a URL, and three CSS-selector strings:
  - item_selector: selects an item (which will get a click-handler attached)
  - item_label_selector: selects a label inside a an item
  - item_value_selector: selects the value inside an item (simply the value which goes into the hidden field, not necessarily a pk)
  I think with that we can basically cover all scenarios and also use any source (which allows CORS requests) with this.
- The search components uses htmx.onLoad (https://htmx.org/docs/#events) when loaded (using ```onload="htmx.onLoad(....)"```) to add a listener for new search results coming in.
- The listener attaches onclick handlers to all items which match item_selector (maybe using ```$$({item_selector})._.bind(...)```?). The onclick handlers should update the label and the value of the component by using item_label_selector and item_value_selector to extract the correct values. Getting e.g. the value would be something like ```f"$({item_value_selector}, this)"``` assuming this is inside a python string

I did not carve it out completely, it is just a draft of how I would remove the coupling between search element and search view.
What do you think? If you think it is too much effort for now I could work on this as well.

wipascal: I think this is a great suggestion!

wipascal: I am also thinking whether it would not be nicer to have the `person_search_views` just return JSON and do the rest inside the component.

bread/layout/components/search_select.py
wipascal: will take care of it. resolving this conversation as we have another one open regarding the bigger issue.

bread/layout/components/search_select.py
saemideluxe: Yes, JSON might be appropriate for this use case. I am not sure if we would want to make a full swing to JSON only though (it requires some decisions to be made and converting JSON to HTML is not always as straight-forward as we would hope it to be), but adding a JSON option will likely become necessary if we grow the project...
The reason I avoided it so far is because if we introduce "client-side templating" (meaning getting JSON results and rendering them into some html on the client side) we need to think about how we implement that well. For a single case some quick-n-dirt javascript to generate html would be okay. But it can become a bit tedious if we take the concept further. And at that point we would want to use a javascript-templating library like mustache or handlebars. And so it can lead down a bit of a rabbit-trail :)
In any case, I woul add the css-selector concept (so that it is possible to use views which were not created with autocomplete in mind) and keep the option (via arguments or as a second autocomplete component) to work with JSON responses.

wipascal:
@saemideluxe PR is ready for a second review

bread/layout/components/search_select.py
```diff
@@ -5,6 +5,10 @@
 from .loading import Loading
 from .tag import Tag
 
+ITEM_SELECTOR = "search_result_item"
+ITEM_LABEL_SELECTOR = "search_result_label"
+ITEM_VALUE_SELECTOR = "search_result_value"
+
```
saemideluxe, line 11: Two things about using selectors within components (in general):
1. They should be generic CSS selectors which can directly be passed to ```document.querySelector()``` or ```$()```. Meaning we should not assue it to be a class or an id or anything more specific than a CSS selector (except there is a good reason for it of course).
2. Secondly they should always be parameterizable by the calling code (except they are only used internally, like most cases of ```htmlgenerator.html_id``` and similar). It might be impossible for the using code to change or set a certain CSS class in some cases. Or we have clashes with multiple elements using the same class for some reason. That is also the reason why we have URL query-parameters parameterizable in many places. This kind of parameterize-ability can save the day :fireworks: 

I would even say that we should not provide default selectors. The point is, the code which instantiates the search-component knows the URL and should also know the structure of the returned result. If we provide default selectors in the search_select component we create an unnecessary dependency between the view code and the search_select code, but the dependency should be more between the search_select "user" and the according view.

So, long speech in short: Could you change the variables to be ```__init__``` parameters and allow any CSS selectors?

The calling code in basxconnect could then e.g. use ```"li"```, ```":scope > :last-child"``` and ```":scope > :first-child"``` for the three values (or something nicer, depending on how you structure the view HTML)

saemideluxe (changes requested):
I think this looks really good now, only add generic parameterization for the selectors.

saemideluxe (approved):

PR #46, opened 2021-06-17
Feature/show report fully

saemideluxe:
Add improvments to the ```reports``` app in order to be able to view the report in the browser and have better defaults (redirection and columns).

### DIFF
--- a/bread/contrib/reports/urls.py
+++ b/bread/contrib/reports/urls.py
@@ -55,7 +55,7 @@ class ReadView(views.ReadView):
         order = self.request.GET.get("ordering")
         if order:
             if order.endswith("__int"):
-                order = order[:-5]
+                order = order[: -len("__int")]
                 qs = qs.order_by(
                     models.functions.Cast(order[1:], models.IntegerField()).desc()
                     if order.startswith("-")
--- a/bread/contrib/reports/urls.py
+++ b/bread/contrib/reports/urls.py
@@ -1,6 +1,7 @@
 import datetime
 
 import htmlgenerator as hg
+from django.db import models
 from django.shortcuts import get_object_or_404
 from django.urls import path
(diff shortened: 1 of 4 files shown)


bread/contrib/reports/urls.py
```diff
@@ -45,14 +47,76 @@ def get_layout(self):
...
-    report = get_object_or_404(Report, pk=report_pk)
+class ReadView(views.ReadView):
+    def get_layout(self):
+
+        # ordering, copied from bread.views.browse.BrowseView.get_queryset
+        qs = self.object.filter.queryset
+        order = self.request.GET.get("ordering")
+        if order:
+            if order.endswith("__int"):
+                order = order[:-5]
```
wipascal, line 58: where does the literal `5` come from?

saemideluxe: :D good question
Stripping the ```"__int"``` from the argument. I could write order[:len("__int")]
In any case: It would be nice to improve this feature in the future (by having a more generic approach) to allow for other sorting criterias as well. But caseless string sort and decimal sort solve 90% of the base cases I think.

saemideluxe: or ```order.rpartition("__int")[0]```

saemideluxe: or ```''.join(order.rsplit('__int', 1))```

saemideluxe: Python has just to many string functions ;)

wipascal: 😄 I like `order[:-len("__int")]`! `rsplit` and `rpartition` are a bit less mainstream than slicing, so I would prefer the slicing.

saemideluxe: Agreed :)

wipascal (approved):

PR #53, opened 2021-06-29
Fix/bulkactions

saemideluxe:
This needed some improvments since quite some time anyways. Should fix the "delete all" problem.


bread/views/browse.py
```diff
@@ -59,14 +73,42 @@ def __init__(self, *args, **kwargs):
...
+                (
+                    Link("delete", label=_("Delete"), icon="trash-can"),
+                    delete,
+                ),
+            ]
+        )
 
     def get_layout(self):
         qs = self.get_queryset()
+        # re-mapping the Links because the url is not supposed to be a real name
```
wipascal, line 94: what do you mean with "a real name"?

bread/views/browse.py
```diff
@@ -138,28 +195,70 @@ def get_queryset(self):
...
-                )
-            if isinstance(column, str):
-                column = _layout.datatable.DataTableColumn(
-                    fieldlabel(self.model, column), hg.C(f"row.{column}")
-                )
 
-            columndefinitions[column.header] = lambda row, column=column: hg.render(
-                hg.BaseElement(column.cell), {"row": row}
+# helper function to export a queryset to excel
+
```
wipascal, line 200: I find these two blank lines confusing

bread/views/browse.py
```diff
@@ -59,14 +73,42 @@ def __init__(self, *args, **kwargs):
...
+            kwargs.get("bulkactions")
+            or self.bulkactions
+            or [
+                (
+                    Link("excel", label=_("Excel"), icon="download"),
+                    lambda request, qs: export(qs, self.columns),
+                ),
+                (
+                    Link("delete", label=_("Delete"), icon="trash-can"),
+                    delete,
```
wipascal, line 87: I thought it was not common in Python to import a function directly.

saemideluxe: IMHO if it is not more than a few names then I don't see a problem. If we have e.g.

    from mymodule import (
        func_1,
        func_2,
        func_3,
        func_3,
        func_3,
        func_3,
        func_3,

wipascal (approved):

PR #54, opened 2021-06-30
Refactor/search select

wipascal:


bread/layout/components/datatable.py
```diff
@@ -227,11 +228,13 @@ def with_toolbar(
...
-                            url=searchurl,
-                            query_urlparameter=query_urlparameter,
+                        Search(
+                            widgetattributes={"autofocus": True},
+                            backend=search.SearchBackendConfig(
+                                url=searchurl, query_parameter=query_urlparameter
+                            ),
                             resultcontainerid=resultcontainerid,
-                            resultcontainer=False,
+                            show_result_container=False,
```
saemideluxe, line 237: Naming this way makes more sense :+1:

bread/layout/components/search.py
```diff
@@ -1,69 +1,112 @@
...
 from .icon import Icon
 from .loading import Loading
 
 
+class SearchBackendConfig(typing.NamedTuple):
+    url: typing.Any
+    result_selector: str = None
+    result_label_selector: str = None
+    result_value_selector: str = None
+    query_parameter: str = "q"
```
saemideluxe, line 15: I would be interested what your reasoning behind extracting a config-object is here. I definitely see the case that we need the search-backend to more configurable.

(Background of why I am asking a bit more is: I am always a bit suspicious of factoring parameter objects out: In the worst case we just changed ```func(param1, param2, param3)``` into ```func(FuncParam(param1, param2, param3))``` which is actually worse because now we first need to understand what a ```FuncParam``` object is. If is just about grouping parameters, then using a prefix is sometimes more readable. And it helps that Python has rather flexible parameter handling.)

bread/layout/components/search.py
```diff
@@ -1,69 +1,112 @@
...
+        widgetattributes = {
             "id": "search__" + hg.html_id(self),
             "_class": "bx--search-input",
             "type": "text",
             "placeholder": placeholder or _("Search"),
             **(widgetattributes or {}),
         }
+        if backend:
+            resultcontainerid = (
+                resultcontainerid or f"search-result-{hg.html_id((self, backend.url))}"
```
saemideluxe, line 47: Can we make sure that ```resultcontainerid``` is handled without the use of ```html_id```? It leads to duplicate html ids if we copy the component inside an inline-formset. I would do it the same way as in search_select.py: ```resultcontainerid = f"search-result-attributes['id']}"```

saemideluxe: Remark: In the case that the Search-component is used standalone it is no problem to use html_id, but if we are inside a FormSet it will be (same is true for SearchSelect).

saemideluxe: Remark: Why it works inside FormSet forms with the widget-id is that Django also generates a widget ID for the template form (the one used for adding new forms) but the ID uses ```__prefix__``` instead of an integer number (as for all other places inside the empty-form-template where the number of the form would occur) which is then later replaced by the (client-side) increased value of the form counter.

saemideluxe (changes requested):
Much clearer, thanks for the refactoring. I would only like to get rid of the ```html_id```.

As a side note: When we start implementing future HTMX/AJAX-style elements (e.g. the new interface), we could try to come up with some way to abstract away some of the JS code by implementing HTMX so that it generates the JS code automatically. That would be really nice (and interesting to implement).

saemideluxe:
Also, removing lines is always a win.

bread/layout/components/search.py
saemideluxe: screenshot-1625114339_246x454
Just to clarify the actual problem: The two inputs (top and bottom) have the same ID, happens when I add two new inline-entries in the realtionship-table.

bread/layout/components/search.py
wipascal: The motivation is exactly the same as for extracting functions.
Also with extracting functions your argument is valid: I agree that changing
`stmt1; 
stmt2; 
stmt3` 
to 
`foo();
def foo():
    stmt1;
    stmt2;
    stmt3;
`
is not a benefit. But if we have something like `stmt1;...stmt14;`, and `stmt4;...stmt8;` are more related to each other than to the rest, it can be an improvement to extract `stmt4;...stmt8;` into a separate method. In my opinion this applies to any kind of data. As soon as tuples, argument lists, etc. reach a certain length, it can be helpful to add more structure.

saemideluxe: I agree that it can be helpful, but as long as the sole purpose is grouping parameters I will stay skeptic... Better naming could be an alternative as well (I think that naming does not get attention it deserve sometimes ;)
Of course when passing config-objects around or having config-generators/shortcuts or similar "consumers" of the additional API then I see a more favorable case for factoring parameters out. And actually I see that to be a good use case for the search backends: Basically that a view could e.g. provide its own SearchBackendConfig.

But anyway, just bringing in my opinion here, no reason to stop merging.

saemideluxe (approved):
Okay, looks good.

PR #55, opened 2021-07-03
Refactor/precompiled css

saemideluxe:
Leaving the compilation of SCSS files in the hand of the package users seems unnecessary in most cases. While we have a django-compressor plugin for automatic compilation, this takes way too long and is only helpfull when working actively with the SCSS files. This PR suggest that we always deliver a precompiled CSS file which should be used by default. For convenience of SCSS development there is a make target which uses ```entr``` to re-compile the CSS when any of the SCSS files change. There are also SCSS/SASS compiler which have a built-in watch function.

This PR makes the initial setup a bit easier and we can get rid of a native depency (libsass) for package users.


saemideluxe:
I don't understand how the "tutorial-skeleton" commit got into this branch... It can be ignored, it is WIP.

wipascal:
Good idea.
I thing we should also remove the `compile_scss` target from the basxconnect_XYZ repos since they are no longer useful.

wipascal (approved):

saemideluxe:
> Good idea.
> I thing we should also remove the `compile_scss` target from the basxconnect_XYZ repos since they are no longer useful.

Right. I will do it for the demo repo and my other projects.

PR #56, opened 2021-07-06
Feature/testing toolkit

saemideluxe:
A first implementation to automatically run model-related tests with hypothesis. I think this is promising, but in order to get the full potential we will need to add more things, parameterization and maybe a coverage analysis.


wipascal (approved):
Cool, thanks!

PR #58, opened 2021-07-13
Feature/persistent tab selection

saemideluxe:
I looke at a few different ways to implement this. URL fragments, URL query parameters and cookies. Cookies are they most sensible option since we handle everything client side and this is a purely client-side state of the interface. I have seen this implemented via URL fragments but that does not work well if we do POST forms and get a redirect because the URL fragment is not sent to the server, so the redirect leads to discarding of the URL fragment and the selected tab is lost. In order to use the cookies I had to introduce a cross-request identifier for the tabs in order to be able to select the correct tab again on the next page load. html_id from htmlgenerator is not stable across requests.
The disavantage is that tab selection will not work as expected for the user if we have tabs with the same labels. However, I think it should not be too much of an issue because keeping the correct tab selected is mostly relevant when working on a single database object. Across different pages it will be less noticeable and might lead in a few cases to an additional click. I though about adding the current URL to the tab-id/cookie key but that would break the tab-selection for the read/edit URLs because they have different URLs. The best option would be to pass a callable to the Tabs __init__ method which can generate a persistent unique ID for a tab and would have the current behaviour as default. That way we could pass e.g. the persons id into the tab-identifier. Maybe I will add a commit for that too.

I though I dump this train of though here because 1. similar might come up in the future with pure client-side UI state and 2. I did write this a bit in a rush (after working too long on feature/allow-tabs-selection-via-query-parameter which is no longer relevant) and might not have thought everything through here ;)


wipascal:
Hmm I don't find it ideal. Is there no way of using part of the url for that? like `/basxconnect/core/naturalperson/read/19/<selected_tab>`

saemideluxe:
Can you explain more what you find it not ideal? Maybe we can work on the suboptimal parts. But as far as I can judge using cookies/local storage is the standard way to do UI state. Thinking about I just realized that I probably should implement the sidebar menu the same way :)

Two reasons why URLs are not a good solution are:
- Using URLs and processing them on the server introduces more dependencies between components. We need to make sure all actions which lead to some kind of reload of the page have the correct URL. For example the View/Edit buttons.
- Tabs can only be changed via page reload. This can be acceptable but if we can avoid it that would be preferable. And while switching tabs during an edit the old tab would need to be saved first.

You can take a look at the branch I mentioned if you want:
https://github.com/basxsoftwareassociation/bread/compare/feature/persistent-tab-selection...feature/allow-tabs-selection-via-query-parameter
The code is not too different but I stopped working on it when I saw how many changes I have to make in the basxconnect repo.

saemideluxe:
And we have the same problem with the cross-request uniqueness of the tab-id...

wipascal:
If it is standard, it's fine for me. I just thought cookies were for saving user-specific stuff like "this user likes cat videos" or "this user prefers the dark theme" etc. (Things that make sense to be persisted for quite long.)

wipascal (approved):

saemideluxe:
Ok. LocalStorage might actually be the cleanest option (basically THE solution for client side persistent state), but I am not too familiar with it.
The only non-cookie solution with cross-request persistence would be a session variable... But then again, it feels wrong to store client side UI-state on the server.

PR #60, opened 2021-07-14
use "display: none" in order for the tag not to use visual space

wipascal:


saemideluxe (approved):

saemideluxe:
Sometimes ```display: initial``` is also an option for re-displaying, but I had cases where it did not work as expectet (how unexpected for CSS ;)

PR #61, opened 2021-07-19
make "lg" the default breakpoint for columns

wipascal:
It just feels awkward to have to type "lg" every time since we only currently target deskop anyway.
This PR **doesn't** break any usages of other breakpoints than "lg".

### DIFF
--- a/bread/layout/components/grid.py
+++ b/bread/layout/components/grid.py
@@ -24,16 +24,12 @@ class Row(htmlgenerator.DIV):
 
 
 class Col(htmlgenerator.DIV):
-    def __init__(self, *children, breakpoint=None, width=None, **attributes):
+    def __init__(self, *children, breakpoint="lg", width=None, **attributes):
         """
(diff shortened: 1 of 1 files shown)


saemideluxe:
At the point where I was using breakpoints, they would not work as expected if the width was given. Does that work now as expected?

wipascal:
> At the point where I was using breakpoints, they would not work as expected if the width was given. Does that work now as expected?

I would say so. I can definitely use width in order to have different widths of columns...

saemideluxe:
Sorry, I meant "if the width was **not** given" ;)

wipascal:
Ah yes, I think then setting a breakpoint does not have any effect.
The default "lg" will only have an effect on the produced html if a width was given.

saemideluxe:
Ah right. I was only concerned for users trying to set the width without a breakpoint then then not getting any results. But with having a default breakpoint this is no longer a concern.

saemideluxe (approved):
Ok

PR #62, opened 2021-07-20
add: view for bulk restore, did: refactoring

saemideluxe:
Refactoring: bulk functions no longer determine what happens after the function finished, the BrowseView is responsible for further processing, e.g. construct the correct redirect.
The reason for this is that the bulk action function are not supposed to know the "calling convention", they only get the queryset. Therefore, in order to construct the correct redirect, the redirect needs to be create in the according BrowseView. I think this is sufficient for now, if we need more we can make the behaviour parameterizable later.


saemideluxe:
Ah, and a small improvment for the bulk-delete function: The success message does no longer count already deleted entries.

PR #63, opened 2021-07-23
Feature/eliminate django template system

saemideluxe:
This a rather big refactoring in order to remove the django templating system from bread and bread applications. The motivation behind is:
- Avoiding to work with two different template system (htmlgenerator and django templates)
- Getting rid of old/unnecessary code
- Having all templates available as python objects which can be modified and generated on the fly
- Beeing consisten with the "component' approach all the way through

Additionaly a small extension to the middleware has been brought in to allow sending state from the client to the server. As previsouly mentioned, UI state should reside on the client. However, when generating the interface on the server, some information is necessary to avoid "lagging" or flickering interface changes after the page has been loaded on the client. One example is the state of the sidebar. In order to send the HTML with the correct classes for expanded or collapsed navigation sidebar, the server needs to know what the current session's setting is. (In the past this has been solved with user preferences, but I think this is not a good solution because the collapse-state of the navigation menu does not exclusively depend on user preferences but also on the device or available space on the screen.) This has been implemented with cookies and the django session object.


bread/layout/base.py
```diff
@@ -1,10 +1,32 @@
...
 from bread.utils import pretty_modelname, resolve_modellookup
 from bread.utils.urls import reverse_model
 
 from ..formatters import format_value
 
+CONTEXT_PROCESSORS = tuple(
+    import_string(path)
+    for path in _builtin_context_processors
+    + tuple(settings.TEMPLATES[0]["OPTIONS"]["context_processors"])
+)
```
saemideluxe, line 16: We read the desired context_processors from the default template settings, which is in 99% of the cases what is relevant. In theory we might not even need the ```TEMPLATE``` settings, but there might be external apps which require at least one template engine to be configured, so we leave that setting as it was.

bread/layout/base.py
```diff
@@ -1,10 +1,32 @@
...
+    def __init__(self, cookiename, value):
+        self.cookiename = cookiename
+        self.value = value
+
+    def resolve(self, context, element):
+        return (
+            context["request"].session["bread-cookies"].get(f"bread-{self.cookiename}")
+            == self.value
+        )
+
```
saemideluxe, line 29: We can use this lazy element (e.g. together with ```hg.If```) in order to render output depending on a certain cookie state. Only ```"bread-"``` prefixed cookies are considered. This is thought to be the standard way to manage client-based state which affects rendering of the page on the server side.

bread/layout/components/notification.py
```diff
@@ -16,7 +16,7 @@
 }
 
 
-class InlineNotification(htmlgenerator.DIV):
+class InlineNotification(hg.DIV):
```
saemideluxe, line 19: Added

bread/layout/components/notification.py
```diff
@@ -76,21 +70,23 @@ def __init__(
...
                     aria_label="close",
-                )
+                ),
             )
+        )
         super().__init__(*children, **attributes)
 
 
-class ToastNotification(htmlgenerator.DIV):
+class ToastNotification(hg.DIV):
```
saemideluxe, line 89: Added auto-fade after n seconds for toast notificaiton (was before only true for toast generated out of django messages).

bread/layout/components/shell_header.py
```diff
@@ -12,27 +13,21 @@ def __init__(self, platform, company, actions=(), *args, **kwargs):
...
-                    )
-                ),
-                hg.A(
-                    hg.SPAN(platform, _class="bx--header__name--prefix"),
-                    hg.SPAN(company, style="position: absolute; left: 17rem"),
-                    _class="bx--header__name",
-                    style="font-weight: 400",  # override carbon design
-                    href="/",
-                ),
+                HasBreadCookieValue("sidenav-hidden", "true"),
```
saemideluxe, line 16: For example: The client called setBreadCookie("sidenav-hidden", "true") and we can check the according value here.

bread/layout/skeleton.py
```diff
@@ -0,0 +1,71 @@
...
+                ),
+                style="position: fixed; right: 0; z-index: 999",
+            ),
+            hg.DIV(*content, _class="bx--content"),
+            hg.SCRIPT(src=_static("js/main.js")),
+            hg.SCRIPT(src=_static("js/bliss.min.js")),
+            hg.SCRIPT(src=_static("js/htmx.min.js")),
+            hg.SCRIPT(src=_static("design/carbon_design/js/carbon-components.js")),
+            hg.SCRIPT(src=_static("djangoql/js/completion.js")),
+            hg.SCRIPT("CarbonComponents.watch(document);"),
```
saemideluxe, line 66: While replacing the original ```base.html``` I had to discard the django-compressor. So we get now a few more requests from the browser to load scripts, but I think we can add our own asset-generator later. The static assets are anyway one of the smaller concerns regarding performance.

bread/middleware.py
```diff
@@ -7,6 +9,11 @@ def __init__(self, get_response):
         self.get_response = get_response
 
     def __call__(self, request):
+        if "bread-cookies" not in request.session:
+            request.session["bread-cookies"] = {}
+        for k, v in request.COOKIES.items():
+            if k.startswith("bread-"):
+                request.session["bread-cookies"][k] = urllib.parse.unquote(v)
```
saemideluxe, line 16: This is the middleware extension for makeing clients-side "bread" cookie available and persistent in the current session.

bread/settings/required.py
```diff
@@ -1,11 +1,24 @@
+"""
```
saemideluxe, line 1: Restructured the settings file and add more comments.

bread/static/js/main.js
```diff
@@ -175,3 +166,16 @@ function submitbulkaction(table, actionurl, method="GET") {
...
+    document.cookie = "bread-" + key + "=" + encodeURIComponent(value) + "; path=/";
+}
+
+function getBreadCookie(key, _default=null) {
+    var ret = document.cookie.split('; ').find(row => row.startsWith("bread-" + key + '='))
+    if(!ret)
+        return _default;
+    ret = ret.split('=')[1];
+    return ret ? decodeURIComponent(ret) : _default;
+}
```
saemideluxe, line 181: These cookie functions can be used to communicate client side changes which affect the generation of the user interface on the server side.

saemideluxe:
This refactroing should be almost completely API compatible, the only thing I had to add in basxconnect was the ```COMPANYNAME``` context variable.

saemideluxe:
Okay, I also had to change views which were relying on ```bread/base.html``` inside basxconnect. So the two wizards.

PR #65, opened 2021-08-04
Feature/adjust paddings

wipascal:


saemideluxe (approved):

saemideluxe:
FYI IMHO this does not require a PR. I think we can maintain some level of "trust" for commiting directly to main and reduce bureaucracy in cases like these. Even if it e.g. breaks the layout I likely would not notice it in the review.
But no problem if you prefer it.

saemideluxe:
Btw sometimes when the CI fails we can just hit "re-run" on the Github Actions page and it will rebuild and also update all dependencies (which was the problem in the first run here I think.

wipascal:
Thanks!
I agree. In cases like these it makes sense to merge right away.

PR #66, opened 2021-08-06
add built-in support for frequently used lazy functions

wipascal:


bread/layout/base.py
```diff
@@ -67,6 +67,26 @@ def resolve(self, context, element):
         return format_value(value)
 
 
+class ObjectFieldLabel(hg.Lazy):
+    def __init__(self, fieldname):
+        self.fieldname = fieldname
+
+    def resolve(self, context, element):
+        return context["object"]._meta.get_field(self.fieldname).verbose_name
```
saemideluxe, line 75: For this one I recommend to take a look at ```fieldlabel``` in the same file. The difference is, that we sometimes want to display a field of a related object, e.g. ```object.primary_email_address.type```. Or we want to display the value of a property on the model but still get a good default label.

bread/layout/base.py
```diff
@@ -67,6 +67,26 @@ def resolve(self, context, element):
...
+        self.fieldname = fieldname
+
+    def resolve(self, context, element):
+        return (
+            getattr(context["object"], f"get_{self.fieldname}_display")()
+            if hasattr(context["object"], f"get_{self.fieldname}_display")
+            else getattr(context["object"], self.fieldname)
+        )
+
+
```
saemideluxe, line 89: Here I would also recommend to take a look at ```formatters.format_value``` (and the recently removed ```render_field``` for in the same file, see in #3445dc). I think we could support looking up any chain of attributes, e.g. with ```htmlgenerator.lazy.resolve_lookup```. So that you could e.g. do ```object.date.isoformat()``` or ```object.primary_email_address.email.upper```.
The code in ```formatters.py``` is likely a bit obsolete as it mixes extracting the value and formatting the value, including producing HTML. The whole formatters idea was significant when we had past Read-Views which did not inherit from EditViews but had their own style. This is getting now more significant again, when work with values and modals.

My suggestion is that we expand ```ObjectFieldValue``` with two things:
- Allow ```fieldname``` to be a chain of dot-accessible attributes, similar to hg.C
- Add an additional parameter ```renderer``` or similar which defaults to ```str``` but could be a custom mapping-function ala signature lambda a: str(a)

What do you think about this? I am sure we will soon encounter cases where such a behaviour is required anyway.

saemideluxe: And then we could also get rid of the file formatters.py, maybe check if we keep some of the default formatter functions.

wipascal: In general I don't like it if we have multiple ways (my new function and the existing formatters.format_value) of doing similar things, so I find it a good idea to unify the whole thing. But it seems like a bigger task, is it ok if I merge this PR and do it in a future PR?

bread/layout/base.py
wipascal: I am not sure I understand what you meant, but I am now calling the `fieldlabel` method.

saemideluxe: Ok, good.

Yeah, sorry, not very good description. Let me explain: If you have something like

```@property
def age(self):
    return today - self.date
```

 as a model method then your initial version would not work when trying to display the label for ```age```, it works exlusively on database fields. But the point is, there are many cases where you want to have a dynamically calculated value (be it via ```queryset.annotate``` or just a method of the Model class) or an attribute of a field's value, not the value directly (that was the primary email address type example).
That is where the ```resolve_modellookup``` inside ````fieldlabel``` tries to get the correct type, callable or value. And then on *that* the ```verbose_name```

bread/layout/base.py
saemideluxe: Yes sure, go ahead. Can you add a TODO comment or something there?

wipascal: Done

saemideluxe (approved):
Thanks.

saemideluxe:
I will merge this for you, I find it good for myself if the code gets already visible on main.

PR #69, opened 2021-08-16
did: remove "Action" and replace with html-links (a href aka Links)

saemideluxe:


wipascal:
In links.py there is a small bug. it should be `List[str]` instead of `list[str]`

wipascal:
very nice, I like it!!

saemideluxe:
> In links.py there is a small bug. it should be `List[str]` instead of `list[str]`

Ups, thought I replaced all of them. Thanks.

wipascal (approved):

PR #71, opened 2021-08-17
add search bar to header

wipascal:


bread/layout/components/shell_header.py
```diff
@@ -80,3 +62,39 @@ def __init__(self, platform, company, actions=(), *args, **kwargs):
...
+        ),
+        hg.SPAN(
+            searchbar or hg.BaseElement(),
+            style=f"position: absolute; left: {searchbar_position}",
+            _class="theme-gray-100",
+        ),
+        hg.A(
+            hg.SPAN(
+                company,
+                style=f"position: absolute; left: {hg.If(searchbar, hg.BaseElement('50 %'), hg.BaseElement('5rem'))}",
```
saemideluxe, line 94: This will not work as expected, because formatted strings in python are not able to render ```hg.BaseElement``` elements correctly (mainly because the string gets interpolated immediately and the ```__str__``` function is used to convert to objects to strings instead of ```hg.render```). I am aware that a notation as used here would be *much* nicer. There is already an open issue to add an API for this in htmlgenerator: https://github.com/basxsoftwareassociation/htmlgenerator/issues/6

So now I am even more motivated to add the feature ;)

saemideluxe: If you have some input on how the implementation or the usage for this feature should look like, feel free to comment on the issue in htmlgenerator.

saemideluxe: screenshot-1629258443_626x50

saemideluxe: Okay, I added a first version for this to htmlgenerator (released already). With the current implementation the implementation would look like this:

```style=hg.format("position: absolute; left: {}", hg.If(searchbar, "50%", "5rem"),```

It is not perfect yet, compared to f-strings. However, it is fully compatible to the python string formatting system. The f-string notation does not really allow for lazy evalutation of its content, it is immediatley evaluated as part of the string literal.
There could be a solution with a custom function like ```hg.f("position: absolute; left: {hg.If(searchbar, '50%', '5rem')}")``` (note the missing f-string specifier) and dynamically evaluates that string when we render. But we would have the problem on how we can capture the scope where the string is defined and pass it to the place where the string is evaluated.

Comment (saemideluxe) [reply bread/layout/components/shell_header.py:94]: Btw. ```hg.If``` works with strings directly. hg.BaseElement is only required if you want to have a list of things.

File: bread/layout/components/shell_header.py [hunk 2]
```diff
@@ -80,3 +62,39 @@ def __init__(self, platform, company, actions=(), *args, **kwargs):
...
+
+
+def variable_size_header_part(platform, company, searchbar, searchbar_position):
+    return hg.BaseElement(
+        hg.A(
+            logo(),
+            platform,
+            _class="bx--header__name",
+            style="font-weight: 400",  # override carbon design
+            href=hg.F(lambda c: c["request"].META["SCRIPT_NAME"] or "/"),
```
saemideluxe, line 84: In case you wondered about this line: The WSGI protocol allows us to "mount" an application on something else than the root of a domain. Se we can e.g. have https://website.com/app1 and https://website.com/app2 and both running with on the same code and seeing themselfes as beeing on the root. All django URL resolving will work correctly as well with this. But when we use our own constructed URLs (like here, because we just want the root-url of this django instance and we don't know the name of that URL) then we have to take care of it.

bread/layout/components/shell_header.py
wipascal: Alright thanks! it should work now

saemideluxe (approved):
Thanks, new search bar looks great.

PR #73, opened 2021-08-31
hide searchbar if user is not authenticated

wipascal:


bread/layout/skeleton.py
```diff
@@ -39,7 +39,11 @@ def default_page_layout(menu, *content):
             ShellHeader(
                 hg.C("PLATFORMNAME"),
                 hg.C("COMPANYNAME"),
-                searchbar=hg.F(lambda c: c.get("SEARCHBAR", "")),
+                searchbar=hg.If(
+                    hg.C("request.user.is_authenticated"),
+                    hg.F(lambda c: c.get("SEARCHBAR", "")),
```
saemideluxe, line 44: I think ```hg.C("SEARCHBAR")``` would work the same and be a bit shorter.

bread/layout/skeleton.py
```diff
@@ -39,7 +39,11 @@ def default_page_layout(menu, *content):
             ShellHeader(
                 hg.C("PLATFORMNAME"),
                 hg.C("COMPANYNAME"),
-                searchbar=hg.F(lambda c: c.get("SEARCHBAR", "")),
+                searchbar=hg.If(
+                    hg.C("request.user.is_authenticated"),
+                    hg.C("SEARCHBAR", ""),
```
saemideluxe, line 44: ```hg.C``` does only support a single argument... Did this not raise an exception? (If the value does not exists ```None``` is returned which is rendered as an empty string (i. e. not rendered).

saemideluxe: Thinking about it: Adding support for a default value could actually make sense in some cases...

wipascal: > `hg.C` does only support a single argument... Did this not raise an exception? (If the value does not exists `None` is returned which is rendered as an empty string (i. e. not rendered).

fixed it, thanks.

saemideluxe (approved):

PR #75, opened 2021-09-04
Refactor/datatable button link improvments

saemideluxe:


saemideluxe:
Sorry for the changes in ```bread/contrib/workflows```, I forgot to switch branches when working on that part.

bread/views/browse.py
```diff
@@ -220,6 +226,18 @@ def get_queryset(self):
                 )
         return qs
 
+    @staticmethod
+    def gen_rowclickaction(modelaction):
```
wipascal, line 230: I would find it nice to use an enum here.. or at least a docstring that lists the possible modelactions

wipascal (approved):
looks very good!

bread/views/browse.py
saemideluxe: Well, we cannot have a definite list of modelactions because they depend on the model-views that have been added and their names. But right, a doc string would make sense to explain the behaviour.

saemideluxe: Okay, added a docstring.

PR #76, opened 2021-09-07
Feature/field information for reports

saemideluxe:
Concern #102


bread/contrib/reports/views.py
```diff
@@ -0,0 +1,265 @@
...
+from bread import layout as layout
+from bread import views
+from bread.utils import filter_fieldlist, generate_excel, xlsxresponse
+from bread.utils.links import ModelHref
+
+from ...layout.components.datatable import DataTableColumn, sortingname_for_column
+from .models import Report
+
+
+class EditView(views.EditView):
```
wipascal, line 21: I almost always find it confusing to have several classes with the same name (`EditView`). I prefer long class names (i.e. `ReportEditView`), but if you prefer it the other way around, I don't mind ;)

saemideluxe: I prefer it this way because I don't like the redundancy in naming, because the file is already in the "report-scope".

saemideluxe: Sidenote:
That kind of preference might also be related to the type of IDE one uses, right? In a full featured IDE context-switches happen much faster in many cases, you just click on the filename or CTRL + MLB whereas in a more commandline-based IDE you have to open files more "consciously" or in a more manual way...

wipascal: I agree it depends on the IDE and general workflow.
But why do you need the "View" in the class name? ;) It's already in the views module!

wipascal (approved):

bread/contrib/reports/views.py
saemideluxe: Correct :D 
It bothers me a little, but at the time I wrote those views I tried to mimick the Django naming scheme a little.
Anyway, in the contrib-modules I am bit less worried about naming and code-style.

PR #78, opened 2021-09-10
add: implementation of a user-profile-page

saemideluxe:
And a few improvments for modals/ajax smuggled in ;)

### DIFF
--- a/bread/layout/base.py
+++ b/bread/layout/base.py
@@ -66,53 +66,28 @@ class FormattedContextValue(hg.ContextValue):
         return str(format_value(super().resolve(context)))
 
 
-# notes on changes (this may be removed py wipascal):
-# By inheriting from ContextValue we can use the parameter
-# "object_contextname" to query any object from the context which
-# is "dot-accessible". E.g we could use ObjectFieldValue("primary_email_address", "row")
-# or ObjectFieldValue("email", "row.primary_email_address")
-# in the second example we "extract the email-address object from the context and
-# use the field "email" of that object.
-#
-# The object can then be obtained from the context by calling super().resolve(context)
-
-
 class ObjectFieldLabel(hg.ContextValue):
-    def __init__(self, fieldname, object_contextname="object"):
+    def __init__(self, fieldname, object_contextname="object", title=True):
+        """
+        :param fieldname: Name of the model field whose value will be rendered
+        :param object_contextname: Name of the context object which provides the field value
+        """
         super().__init__(object_contextname)
         self.fieldname = fieldname
+        self.title = title
 
     def resolve(self, context):
-        return fieldlabel(super().resolve(context)._meta.model, self.fieldname)
-
-
-# notes on changes (this may be removed py wipascal):
-# The additional complexity comes from a use case where
-# we want to display a value which spans one or multiple relationships
-# E.g. if we have a table of persons and the object context-name is
-# "row" and we want to display the country, we would want to specifiy the field
-# as "primary_postal_address.country". This is what is already mentioned in the
-# first note above, but we make use of hg.resolve_lookup instead of fieldlabel (
-# because resolve_lookup is much more generic, while fieldlabel only works with
-# django models and has therefore a bit more information when doing the lookup).
-
-# If "primary_postal_address.country" would have a "get_country_display" method,
-# then we want to call that method as "primary_postal_address.get_country_display".
-# this is why we do the slightly ugly string-splitting
-
-# Note that we might also want to access a non-django-field value like
-# "primary_postal_address.country.name" which works with the implementation below
-
-# I think it is good if we have a standard way to display database values and field
-# labels, so I am all for replacing the existing methods with a consistent one.
-# The amount of "magic" or "genericnes" that I would like to have here is because
-# it makes prototyping much faster. The dot-lookup-anything idea comes from
-# traditional html-template systems like the django template language, hg.resolve_lookup
-# is copied and adjust from the django template system.
+        ret = fieldlabel(super().resolve(context)._meta.model, self.fieldname)
+        return ret.title() if self.title else ret
 
 
 class ObjectFieldValue(hg.ContextValue):
     def __init__(self, fieldname, object_contextname="object", formatter=None):
+        """
+        :param fieldname: Name of the model field whose value will be rendered
+        :param object_contextname: Name of the context object which provides the field value
+        :param formatter: function which takes the field value as a single argument and returns a formatted version
+        """
         super().__init__(object_contextname)
         self.fieldname = fieldname
         self.formatter = formatter
--- a/bread/layout/base.py
+++ b/bread/layout/base.py
@@ -19,7 +19,9 @@ class HasBreadCookieValue(hg.Lazy):
         self.default = default
 
     def resolve(self, context):
-        if f"bread-{self.cookiename}" in context["request"].session["bread-cookies"]:
+        if f"bread-{self.cookiename}" in context["request"].session.get(
+            "bread-cookies", {}
+        ):
             return (
                 context["request"].session["bread-cookies"][f"bread-{self.cookiename}"]
                 == self.value
--- a/bread/layout/base.py
+++ b/bread/layout/base.py
@@ -9,6 +9,8 @@ from bread.utils.urls import reverse_model
 
 from ..formatters import format_value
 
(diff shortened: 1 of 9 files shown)


bread/layout/base.py
```diff
@@ -66,53 +77,28 @@ def resolve(self, context):
...
-# or ObjectFieldValue("email", "row.primary_email_address")
-# in the second example we "extract the email-address object from the context and
-# use the field "email" of that object.
-#
-# The object can then be obtained from the context by calling super().resolve(context)
-
-
 class ObjectFieldLabel(hg.ContextValue):
-    def __init__(self, fieldname, object_contextname="object"):
+    def __init__(self, fieldname, object_contextname="object", title=True):
```
wipascal, line 81: Hmm the introduction of `object_contextname` seems to extend the functionality in a awkwardly limited way. Maybe we can just support dots in the fieldname so that arbitrary object names and nestings are possible...?

bread/layout/components/modal.py
```diff
@@ -81,6 +81,13 @@ def __init__(
             **attributes,
         )
 
+    def attach_to_trigger(self, trigger: hg.BaseElement):
+        """
+        Attaches this modal to a trigger element and returns a wrapper element
+        """
+        trigger.attributes.update(self.openerattributes)
```
wipascal, line 88: For me, attribute updates (especially dicts) are never a good thing. I find immutable objects much easier to reason about. 
Therefore I would prefer a shortcut function that takes everything that is needed to create the modal _and_ the trigger and creates both. In this specific case the overhead that is needed seems significant, but I think immutability wins in the long run.

bread/views/userprofile.py
```diff
@@ -0,0 +1,268 @@
...
+
+    def check_permissions(self, request):
+        if not request.user.is_superuser:
+            return HttpResponseForbidden()
+
+    def get_object(self):
+        return self.request.user
+
+
+def profile_field(fieldname, password=False):
```
wipascal, line 217: I would create two functions "profile_regular_field(fieldname)" and "profile_password_field(fieldname)" which both call a function "profile_field(fieldname, fieldvalue)".
For me, a function should focus as much as possible to do one thing instead of using if-else.

bread/layout/components/modal.py
wipascal: btw this shortcut function could also take care of adding the "asajax" parameter and maybe other stuff.

bread/layout/base.py
saemideluxe: Regarding the limitation: The name of the object can be dotted as well as long as the latest member of the dot-chain is a Django Model instance. And the fieldname can be dotted as well, but here we just try to be smart about that last members "verbose_name". So I think that is not a big concern.

The problem with a single dotted value would be that it would not be clear in all cases which part of the dot-chain represents the model instance. E.g. ```object.name``` it would be the first element, in ```request.user.email``` it would be the second element and in ```object.primary_adress.type.name``` it would be the third one.

But I agree that there is a bit of a problem here with the implicit assumption of ```object``` beeing in the context by default. This makes sense for Djangos class-based read and edit views, but not in other parts of the interface. A single string would make it a nicer API.

So, if we think having a single string (to declare the object and the label) is imporant there might be a solution, but it would not be trivial: We would need to resolve from "end-to-start" in order to determine the first member that is a Django model instance. But in order to go from end to start we need to look up pairwise because we cannot reverse-access a python object in general.
I think I can give it a short try, but if it doesn't work I will put in on lower priority, okay?

bread/layout/components/modal.py
wipascal: We are using the "construct" of modal with a trigger in a number of places now. And still it is not simple to create even the most trivial working instance of this construct. I think we should make this very easy. So maybe we can just have a shortcut "modal_with_trigger(path_of_view: str, [optional params])" that creates a "edit"-button and a modal with the title "edit" and a lot of optional parameters to customize where needed.

bread/layout/base.py
wipascal: I thought the fieldname never has dots. If it makes sense to allow dots in the fieldname, my suggestion is not relevant.

bread/layout/components/modal.py
wipascal: the title of the modal can also be more clever than just "edit", it could be "Edit <automatically figure out name of the model>"

saemideluxe: I am not sure what exactly you would like to change. You would create a new dict and assign it to ```trigger.attributes```? Then we still mutated the trigger element, so we still have state change. Yes, we could avoid to change the state of ```attributes``` but it wouldnt' make sense because dict is a mutable object.

Maybe some background here (because ```attributes``` is coming from htmlgenerator): I decided to write htmlgenerator with mutable datastructures. I did some experiments with immutable datastructures first. But it turned out that the code would in many cases be more complicated to write, especially the API and construction of tree elements. Not necessarily because of the mutability per se but because Python is in most cases designed around the assumption of mutable objects.

I agree that immutability can be an advantage in many cases. But I would argue that it should not be imperative just for the sake of it. The reason for my thinking is: Mathematics is immutable but the real world is mutable. Therefore we will not be able to "factor out immutability" in many cases.

bread/views/userprofile.py
saemideluxe: Agreed, this is unclean.

bread/layout/components/modal.py
wipascal: > I am not sure what exactly you would like to change. You would create a new dict and assign it to trigger.attributes?

No, that's not what I meant. I just prefer it if the trigger is created with the right attributes from the beginning. Assigning a new dict would be just as ugly and wouldn't help at all because no one holds a "pointer" to the dict anyway.

So I am actually just saying that I in principle prefer it the way we are already doing it at other places, but I agree that we should streamling the process of creating triggers and button and I am proposing another way of doing this.

bread/views/userprofile.py
```diff
@@ -0,0 +1,268 @@
...
+                                reverse("userprofile.personal", query={"asajax": True}),
+                                submitlabel=_("Save"),
+                            ).attach_to_trigger(
+                                layout.button.Button(
+                                    _("Edit"),
+                                    buttontype="tertiary",
+                                    icon="edit",
+                                    style="margin-top: 1rem",
+                                )
+                            ),
```
wipascal, line 56:
```suggestion
                            modal_with_button(
                                title=_("Personal Information"),
                                view_path="userprofile.personal",
                                submitlabel=_("Save"),
                                button_("Edit"),
                                buttontype="tertiary",
                                buttonicon="edit",
                                button-style="margin-top: 1rem",
                            ),
```

wipascal (approved):
Looks good.

bread/views/userprofile.py
wipascal, line 56:
```suggestion
                            ).attach_to_trigger(
                                lambda attributes: layout.button.Button(
                                    _("Edit"),
                                    buttontype="tertiary",
                                    icon="edit",
                                    style="margin-top: 1rem",
                                    **attributes
                                )
                            ),
```

bread/urls.py
```diff
@@ -86,6 +88,16 @@ def get_success_message(self, data):
     path("auth/", include("django.contrib.auth.urls")),
     path("accounts/login/", auth.BreadLoginView.as_view(), name="login"),
     path("accounts/logout/", auth.BreadLogoutView.as_view(), name="logout"),
+    path("accounts/user/", userprofile.UserProfileView.as_view(), name="userprofile"),
+    autopath(
+        userprofile.EditPersonalDataView.as_view(), urlname="userprofile.personal"
```
wipascal, line 93: I was just wondering: Would you find it ugly to have this urlname as a static method/variable in the EditPersonalDataView? This would help prevent typos...

saemideluxe: In short: I think it is simply not necessary.

In long: In case a user needs to use the view he either needs to lookup this hardcoded string identifier or the name of the static variable. So it requires scaning the code anyway to get the correct identifier (be it string or variable) and the typo can happen with a variable name all the same.

Also, we have a good tool for finding names of URLS: ```manage.py show_urls```. This is what I use most often do discover URLs of other apps.

If I want to avoid hardcoded strings for URL names I use autopath without the ```urlname``` parameter.

PR #79, opened 2021-09-13
Feature/rowaction modals

wipascal:
I know you were against the lambda parameter for the shortcut...
But in this case I didn't find a better solution. I am open for suggestions.

### DIFF
--- a/bread/contrib/reports/migrations/0005_auto_20210826_1951.py
+++ b/bread/contrib/reports/migrations/0005_auto_20210826_1951.py
@@ -6,18 +6,22 @@ from django.db import migrations, models
 class Migration(migrations.Migration):
 
     dependencies = [
-        ('reports', '0004_auto_20210619_1502'),
+        ("reports", "0004_auto_20210619_1502"),
     ]
 
     operations = [
         migrations.AlterField(
-            model_name='report',
-            name='id',
-            field=models.BigAutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID'),
+            model_name="report",
+            name="id",
+            field=models.BigAutoField(
+                auto_created=True, primary_key=True, serialize=False, verbose_name="ID"
+            ),
         ),
         migrations.AlterField(
-            model_name='reportcolumn',
-            name='id',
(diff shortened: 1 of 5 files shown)


saemideluxe:
Right, row-click-modal issue :D 

In this PR I find more the additional attribute ```Link.modal``` problematic. It is changing the Link API in a very specific way for a rather limited use case. For such cases my idea (and recent change) was to have the additional ```attributes``` on the Link class to specify special behaviour via HTML attributes.

I would also treat the row-click-modal issue as "experimental" until we learn a bit more how and where we use it. What I mean with that is that I would not yet make permanent changes inside bread itself, especially when we need to extend an API.

But I see the problem here, I tried to implement the modal myself and for a clean implementation there was at least something missing: The ```id``` of the modal could not be set dynamically. I fixed that, because ```id``` is a normal HTML attribute and should therefore be allowed to use ```hg.Lazy``` value.
Now, with that we can make a more or less clean implementation without changing bread core:



```python
class CaBrowseView(BrowseView):
    # define the modal with lazy values only and a custom id
    rowedit_modal = layout.modal.Modal.with_ajax_content(
        "Edit",
        ModelHref(
            certmodels.Ca,
            "edit",
            kwargs={"pk": hg.C("row.pk")},
            query={"asajax": True},
        ),
        submitlabel="Save",
        id=hg.format("row-modal-{}", hg.C("row.pk")),
    )
    # add an empty column for the modal
    # the modal could also be attached to an existing column
    # instead of creating a new one
    columns = [
        "name",
        "key_length",
        "digest",
        "created",
        "modified",
        layout.datatable.DataTableColumn("", rowedit_modal),
    ]
    # set the row action
    # it is mostly empty and only needs the openerattributes
    # onclick and onauxclick are set to none for aesthetic reasons ;)
    rowclickaction = Link(
        "#",
        "",
        attributes={
            **rowedit_modal.openerattributes,
            "onclick": None,
            "onauxclick": None,
        },
    )

```

saemideluxe:
Adding a shortcut at some point might be usefull though...

PR #85, opened 2021-11-02
add checks to disable buttons who are coming from links without enough permissions

saemideluxe:


wipascal (approved):

wipascal:
would be nice to fix the CI 😄

saemideluxe:
Right. I just fixed this earlier today, but on main ;)

PR #86, opened 2021-11-02
Bugix/miscellaneous

wipascal:


bread/views/browse.py
```diff
@@ -1,4 +1,4 @@
-from typing import Callable, List, NamedTuple, Optional, Union
+from typing import Callable, List, NamedTuple, Optional, Tuple, Union
```
saemideluxe, line 1: This is a workaround because Python < 3.9 does not allow to subscribe to ```tuple```. Can be reverted around 2023 ;)

bread/layout/base.py
```diff
@@ -116,7 +116,8 @@ def resolve(self, context):
         parts = self.fieldname.split(".")
         # test if the value has a matching get_FIELDNAME_display function
         value = hg.resolve_lookup(
-            object, f"{'.'.join(parts[:-1])}.get_{parts[-1]}_display"
+            object,
+            f"{''.join([part + '.' for part in parts[:-1]])}get_{parts[-1]}_display",
```
saemideluxe, line 120: Can you quickly explain what this change improves/fixes?

wipascal: I had explained it in the commit message ;)
if there is only one part we ended up with an unwanted leading dot (e.g. ".get_FIELDNAME_display")

saemideluxe: I see, missed the commit message. This could also be solved with
```python
f"{'.'.join(parts[:-1])}.get_{parts[-1]}_display".lstrip(".")
```
which I slightly prefer, but it is fine either way.

saemideluxe (approved):

PR #87, opened 2021-11-02
fix: remove the use of get_concrete_instance

saemideluxe:
This is not necessary in many places, instead the newly added
InheritanceManager should be used since it will fetch all concrete
instances in a single query.

### DIFF
--- a/bread/formatters.py
+++ b/bread/formatters.py
@@ -6,7 +6,6 @@ import htmlgenerator as hg
 from dateutil import tz
 from django.conf import settings
 from django.db import models
-from django.urls import NoReverseMatch
 from django.utils.functional import Promise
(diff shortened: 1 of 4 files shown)


saemideluxe:
get_concrete_instance is still in use in some places, but the most concerning one is the use of it browse-views. Single uses less serious and in case of ```bread.utils.urls.reverse_model``` quite usefull. But maybe we should remove it even there, just to prevent accidental use of it.

wipascal:
Hmm in this PR I don't see a usage of the code in inheritancemanager.py 
How to use it?

saemideluxe:
See here:
https://github.com/basxsoftwareassociation/basxconnect/pull/173

wipascal (approved):

PR #88, opened 2021-11-03
Fix/userprofile permission

dward2nd:
I fixed the problem that normal users cannot access their own profile view.


saemideluxe (approved):
Thanks.

PR #94, opened 2021-11-08
Feature/views improvements

saemideluxe:


bread/formatters.py
```diff
@@ -9,41 +9,36 @@
...
     if isinstance(value, models.fields.files.FieldFile):
         return as_download(value)
     if isinstance(value, Iterable) and not isinstance(value, (str, bytes, Promise)):
         return as_list(value)
+    if isinstance(value, str):
+        if "\n" in value:
+            return as_text(value)
+        if is_email_simple(value):
+            return as_email(value)
+
```
saemideluxe, line 41: Since we make the formatter independent of the model field, we need to add a bit of output-sugar manually.

bread/formatters.py
```diff
@@ -149,19 +137,26 @@ def as_video(value):
...
-
 CONSTANTS = {
     None: getattr(settings, "HTML_NONE", app_settings.HTML_NONE),
     True: getattr(settings, "HTML_TRUE", app_settings.HTML_TRUE),
     False: getattr(settings, "HTML_FALSE", app_settings.HTML_FALSE),
 }
+
+
+# copied from https://github.com/django/django/blob/1f9874d4ca3e7376036646aedf6ac3060f22fd69/django/utils/html.py
+# because older versions only have this as a local function available
```
saemideluxe, line 148: Older versions == Django 3.2 does not have this function accessible from outside of the urlize funtion.

bread/layout/base.py
```diff
@@ -117,11 +117,12 @@ def resolve(self, context):
...
             value = hg.resolve_lookup(object, self.fieldname)
-        if self.formatter:
-            return self.formatter(value)
         if isinstance(value, datetime.datetime):
             value = localtime(value)
-        return localize(value, use_l10n=settings.USE_L10N)
+        value = localize(value, use_l10n=settings.USE_L10N)
+        if self.formatter:
+            return self.formatter(value)
+        return value
```
saemideluxe, line 125: datetimes should always be localized, the formatter should get a localized value.

bread/views/read.py
```diff
@@ -45,6 +50,48 @@ def get_required_permissions(self, request):
...
+                    )
+                    for field in self.fields
+                ),
+            ),
+        )
+
+    def get_required_permissions(self, request):
+        return [f"{self.model._meta.app_label}.view_{self.model.__name__.lower()}"]
+
+
```
saemideluxe, line 94: This seems actually a much better option, especially for prototyping than using the readonly-form. It is also better readable.

wipascal (approved):
nice!

PR #96, opened 2021-11-09
extract a function from ObjectFieldValue

wipascal:
I have a usecase for this in an upcoming PR.
(edit: this is the PR https://github.com/basxsoftwareassociation/ponds_next/pull/2)


saemideluxe:
Hmmm... I see why you need something like this. But I would prefer if we could keep the API to a single class or function for rendering values.
We could solve this by makeing the ObjectFieldValue slightly more generic.

```python
object = hg.resolve_lazy(super().resolve(context))
```

Then we can pass in any lazy value.

wipascal:
I don't fully understand what you mean. Would `ObjectFieldValue` then extend `Lazy` instead of `ContextFunction`?

saemideluxe:
Not necessarily but I guess it would make sense in order keep the code easier to understand.

In that case you would not call ```super*(```, but still need to run the code from ```ContextValue.resolve```. I think violating the DYI principle is totally justified here. ```ObjectFieldValue``` started as a simple "extensions" of ```ContextValue``` but now it is evolving into a a more generic value-formatting tool.

saemideluxe:
Das wäre etwa, was ich vorschlagen würde (parameter-name und doc-string muss ich noch anpassen).

wipascal:
> Das wäre etwa, was ich vorschlagen würde (parameter-name und doc-string muss ich noch anpassen).

passt, ist definitiv besser als so wie ich gedacht hatte das du es möchtest.

saemideluxe:
Okay, great (ups, switched to german before :D)

saemideluxe (approved):

saemideluxe (approved):
Okay, thanks for the polish.

PR #98, opened 2021-11-12
Added the testcase to check the visibility from non-logged in users

dward2nd:


bread/tests/test_public_visible.py
```diff
@@ -0,0 +1,99 @@
...
+        "delete",
+        "edit",
+        "read",
+    }
+
+    # (model_objects, url, additional_operations)
+    BREAD_MODELS_URLPATTERNS = (
+        (Report.objects.all(), "/reports/reports/report/", {"excel"}),
+        (TaskResult.objects.all(), "/bread/django_celery_results/taskresult/", set()),
+    )
```
saemideluxe, line 33: This is no longer used, right?

bread/tests/test_public_visible.py
```diff
@@ -0,0 +1,99 @@
...
+        "/bread/accounts/password_reset/",
+        "/bread/accounts/password_reset/done/",
+        "/bread/accounts/reset/<uidb64>/<token>/",
+        "/bread/accounts/reset/done/",
+        "/bread/auth/password_reset/",
+        "/bread/auth/reset/<uidb64>/<token>/",  # password reset
+        "/bread/auth/password_reset/done/",
+        "/bread/auth/reset/done/",
+        # TODO: figure out the way to manage this view.
+        "/bread/preferences/global/<slug:section>",
```
saemideluxe, line 46: Here I would prefere if we use the URL name instead of a hard-coded path. This allows us to change the URL structure in the future without needing to update this test.

bread/tests/test_public_visible.py
```diff
@@ -0,0 +1,99 @@
...
+
+        # update the exceptions for some pages that can be visible to the public
+        # and those pages with dynamic url regex
+        # and those out of the bread's scope (e.g., basxconnect)
+        url_names = self.ALL_URL_NAMES - excepted_urlpatterns - url_intpk
+
+        # manually add a list of urls with a dynamic subdirectory.
+        # (e.g., <int:pk>)
+        for url in url_intpk:
+            url_names.add(url.replace("<int:pk>", str(secrets.randbelow(100))))
```
saemideluxe, line 62: Can you generalize this a bit? I mean so that it works for any type of parameter. If you look at https://github.com/basxsoftwareassociation/bread/blob/main/bread/utils/urls.py#L274 you can see the possible type of parameters. It is also possible to have a parameter without a type, in that case the type is str.
We should also be able to handle any parameter name, not only "pk".
This will also solve the problem with the ```"/bread/preferences/global/<slug:section>"``` URL.

bread/tests/test_public_visible.py
```diff
@@ -0,0 +1,99 @@
...
+
+    def test_visible(self):
+        # add the uncompiled url to exception
+        excepted_urlpatterns = self.EXCEPTED_URLPATTERNS.copy()
+        url_intpk = {x for x in self.ALL_URL_NAMES if "<int:pk>" in x}
+
+        # update the exceptions for some pages that can be visible to the public
+        # and those pages with dynamic url regex
+        # and those out of the bread's scope (e.g., basxconnect)
+        url_names = self.ALL_URL_NAMES - excepted_urlpatterns - url_intpk
```
saemideluxe, line 57: Set operations are a good thing :+1:

bread/tests/test_public_visible.py
```diff
@@ -0,0 +1,99 @@
...
+        # try fetching each url and see the response
+        c = Client()
+
+        for url in url_names:
+            try:
+                response = c.get(url)
+                # if the status code is 4,
+                # there's some errors loading that pages
+                self.assertTrue(
+                    response.status_code // 100 != 4,
```
saemideluxe, line 73: Okay, this is a smart way to solve it ;)
But in Python it would be a bit more readable to write something like 400 ```not 400 <= response.status_code <= 499```

bread/tests/test_public_visible.py
```diff
@@ -0,0 +1,99 @@
...
+                    "This page %s failed to load, with status code %d."
+                    % (url, response.status_code),
+                )
+                # if the status code is 3, and the Location is /bread/accounts/login/
+                # it's likely invisible to anonymous users, as the page
+                # redirects to the login page.
+                self.assertTrue(
+                    response.status_code // 100 == 3
```
saemideluxe, line 81: Same as above

bread/tests/test_public_visible.py
```diff
@@ -0,0 +1,99 @@
...
+                    % (url, response.status_code),
+                )
+
+            except TemplateDoesNotExist as e:
+                if hasattr(e, "message"):
+                    print("The template for", url, "at", e.message, "does not exist.")
+                else:
+                    print("The template for", url, "does not exist.")
+
+                traceback.print_exc()
```
saemideluxe, line 95: Hmmm... Where is this exception coming from? Can you explain why we need to catch it?

bread/tests/test_public_visible.py
```diff
@@ -0,0 +1,99 @@
...
+
+        print(
+            "Public Visibility Test for Bread completed. No confidential pages appears publicly."
+        )
```
saemideluxe, line 99: Tests should normally only have output for errors or warning. So let us remove this.

bread/tests/test_public_visible.py
```diff
@@ -0,0 +1,99 @@
...
+                    "This page %s failed to load, with status code %d."
+                    % (url, response.status_code),
+                )
+                # if the status code is 3, and the Location is /bread/accounts/login/
+                # it's likely invisible to anonymous users, as the page
+                # redirects to the login page.
+                self.assertTrue(
+                    response.status_code // 100 == 3
+                    and response.headers["Location"].startswith(
+                        "/bread/accounts/login/"
```
saemideluxe, line 83: Here we should also not use a hardcoded URL to check against.

bread/tests/test_public_visible.py
```diff
@@ -0,0 +1,99 @@
...
+                    response.status_code // 100 != 4,
+                    "This page %s failed to load, with status code %d."
+                    % (url, response.status_code),
+                )
+                # if the status code is 3, and the Location is /bread/accounts/login/
+                # it's likely invisible to anonymous users, as the page
+                # redirects to the login page.
+                self.assertTrue(
+                    response.status_code // 100 == 3
+                    and response.headers["Location"].startswith(
```
saemideluxe, line 82: The ```headers``` attribute is not always available. Maybe a better way to do the testing would be to do something like ```isinstance(response, HttpResponseRedirect)``` and then check ```response.url```. A good way to test that value would be to use ```django.urls.resolve``` and to see if it resolves to the login page.

saemideluxe (changes requested):
Thanks for the code. I added quite a few comments, but don't worry. This type of testing requires quite a bit of understanding of Django internals, because we try to automate as much as possible.

You can read my feedback below. For easier understanding of the code I would add two helper functions: One to generate a "real" URL from a url pattern and one which checks if a response is correct or not.

bread/tests/test_public_visible.py
dward2nd: Yes, I forgot to remove them.

bread/tests/test_public_visible.py
dward2nd: For that case, both views from `accounts` and `auth` share the same names. Would it be OK if I set the namespace for Django's `auth` to `django_auth` instead, to prevent naming conflict and further difficulties?

bread/tests/test_public_visible.py
dward2nd: There's one exception raised while testing.

```
django.template.exceptions.TemplateDoesNotExist: registration/login.html
```

bread/tests/test_public_visible.py
saemideluxe: The django auth urls should already be inside the "admin" namespace, e.g. ```"admin:login"```. Does this conflict with bread?

saemideluxe: If necessary we could put the bread URLs inside a bread namespace...

bread/tests/test_public_visible.py
saemideluxe: Hmm... How many of these do we have?

If it is only from login or auth views we can catch the exception and silently drop them.

bread/tests/test_public_visible.py
dward2nd: There are views that registered as the same name within `bread`. For example, `/bread/auth/password_reset/` and `/bread/accounts/password_reset/` use `password_reset`, with no any namespaces specified.

bread/tests/test_public_visible.py
dward2nd: Only one here for now.

bread/tests/test_public_visible.py
dward2nd: `/bread/auth/password_reset/` is the one from `django.contrib.auth.views.PasswordResetView`, and `bread.views.auth.BreadPasswordResetView` is the one from `bread.views.auth.BreadPasswordResetView`.

saemideluxe: Ah, I see. Okay, we actually do not need the django-auth views anymore, we replaced everything (except the password-change, but we handle that with password-reset). You can remove ```path("auth/", include("django.contrib.auth.urls"))``` from urls.

bread/tests/test_public_visible.py
saemideluxe: Okay, then I suggest we add an exception for that specific URL. But other URLs should still raise the TemplateDoesNotExist exception.

bread/tests/test_public_visible.py
dward2nd: This one is almost completed. The only problem here is for this URL `/bread/accounts/reset/<uidb64>/<token>/`. I still have no idea how I will reverse and get the exact same path with `<` and `>` because it will be converted to the URL encoded URL.

bread/tests/test_public_visible.py
dward2nd: I replaced all types except `int` with randomized strings. Would you suggest what I should be specifically putting into `UUID`, `slug`, `path`, and even `str`?

saemideluxe: For UUID you should generate a uuid with pythons ```uuid``` module. Otherwise I think djangos URL-parse will raise an error. For slug, path and str a random str should be okay as long as they are alphanumeric.

bread/tests/test_public_visible.py
saemideluxe: Did you try something like this?

```python
reverse("password_reset_confirm", kwargs={"uidb64": "hello", "toke": "world"})
```

bread/tests/test_public_visible.py
```diff
@@ -0,0 +1,96 @@
...
+import django.urls
+from django.contrib.admindocs.views import simplify_regex
+from django.http import HttpResponseRedirect
+from django.template import TemplateDoesNotExist
+from django.test import Client, TestCase
+from django.urls import reverse_lazy as reverse
+from django_extensions.management.commands import show_urls
+
+ROOT_URLPATTERNS = django.urls.get_resolver(None).url_patterns
+ALPHANUMERIC_STR = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789"
```
saemideluxe, line 13: This is fine like this.
If you do not want to type it by yourself every time there is also the following:

```python
import string
ALPHANUMERIC_STR = string.ascii_letters + string.digits
```

dward2nd: That's a good idea. Thank you.

bread/tests/test_public_visible.py
saemideluxe: Okay I see. Personally I would in general try to call ```reverse``` as late as possible and work with url-names as long as possible.

For example (just pseudo-code):

```python
for urlname, urlpattern in ALL_URL_NAMES: # need to adjust creating of ALL_URL_NAMES a bit to include urlnames
    if urlname not in EXCEPTED_URLPATTERNS: # can put only urlnames inside EXCEPTED_URLPATTERNS without reverse
        urls.append(generate_url(urlname, urlpattern)) # call reverse inside generate_url
```
That way you can work with a single for-loop and you do not need to reverse the excepted URLs.

What do you think?

dward2nd: Interesting. I solved this one according to the solution you had suggested using a dictionary, URL names as keys and URL patterns as values.

dward2nd: I think this can be marked as resolved now.

saemideluxe (commented):
Okay, looks good.

saemideluxe (approved):

PR #99, opened 2021-11-12
support displaying the reverse relationship of foreign key

wipascal:
I'd have preferred checking for instance of "RelatedManager", but the problem is that this class is not public so it can't be imported :D

### DIFF
--- a/bread/layout/base.py
+++ b/bread/layout/base.py
@@ -2,6 +2,7 @@ import datetime
 
 import htmlgenerator as hg
 from django.conf import settings
+from django.db import models
 from django.db.models import ManyToOneRel
 from django.http import HttpResponse
 from django.template.context import _builtin_context_processors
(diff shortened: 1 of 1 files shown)


bread/layout/base.py
```diff
@@ -131,6 +134,8 @@ def resolve(self, context):
         value = localize(value, use_l10n=settings.USE_L10N)
         if self.formatter:
             value = self.formatter(value)
+        if hasattr(value, "all"):
+            value = ", ".join(value.all())
```
saemideluxe, line 138: Could we use the same as ```bread.formatters.format_value``` for this?

```python
    if isinstance(value, models.Manager):
        value = value.all()
```

saemideluxe (changes requested):
I saw this functionality was missing at quite a few occasions already, thanks.

saemideluxe (approved):
Cool, thanks.

PR #103, opened 2021-11-30
convert \n into <br> in ObjectFieldValue

wipascal:


bread/layout/base.py
```diff
@@ -137,6 +138,8 @@ def resolve(self, context):
             value = self.formatter(value)
         if isinstance(value, models.Manager):
             value = ", ".join([str(x) for x in value.all()])
+        if isinstance(value, str):
+            value = linebreaksbr(value, autoescape=False)
```
wipascal, line 142: Actually `autoescape=True` seems to lead to exactly the same result... Not sure which one makes more sense.

saemideluxe: Okay, then we can leave it with the default I guess.

saemideluxe (approved):

wipascal:
sorry I just committed another small thing here. should have made another branch

saemideluxe (approved):
Sure

PR #106, opened 2021-12-08
make search select useable in a regular form

wipascal:


saemideluxe (approved):
Approved without a closer look as this will be factored out in the current form-componentes refactoring ;)

PR #109, opened 2021-12-16
Refactor/forms api

saemideluxe:
Work in progress


saemideluxe:
Okay, this is quite a big one. However, in a lot of places there happend just small changes due to how the API changed. The relevant changes are mainly in ```bread.layout.components.forms```. That package is not perfect yet, I see quite a few improvments on how we could restructure it. But we have to get this refactoring through at some point.

The main points of this refactoring were:
- Use better, more consistent names. Especially what formerly was called "FormField" is actually considered a "Widget" in the Django world, widgets are responsible to render a form field, the formfield itself does not have "knowledge" about representation. This is important for us to realize, as we normally only want to modify widgets (to be "carbonized" or similar) and not the handling of fields. This attitude is now much better reflected in the naming and structure.
- Consistent API of how widgets are created/instantiated. This could likely be written in a better way, but we have now a good place to start from. All widget __init__ methods accept the same minimal set of parameters.
- Widgets can now be better instantiated independent from Django forms, directly from their class or via a function (```generate_widget_element```). Normally the later should be used, as it will create a fully "carbonized" widget. However, sometimes the "pure" version of the widget is desired.
- A function (```generate_widget_element```) which can be used to correctly generate a widget that depends on a Django form field. The function will take care to pass all relevant settings from the Django form field to the widget. The function tries to mimick the vanilla Django widget instantiation as closely as possible.
- Add some things which have been missing since quite some time like: Clearable file input and radio selects

saemideluxe:
Also: I added a preview page for widgets. I think we will write some additions there to show the widgets in different states, like with label, error message, disabled, etc.

saemideluxe:
Also also I think we should test this on connect.basx.ch before merging. We can install directly from this branch. And then let a few people test different areas.

saemideluxe:
Also also also, I did not yet test this with UM and Ponds, they will likely need a few small adjustments.

bread/layout/components/forms/fields.py
```diff
@@ -0,0 +1,189 @@
...
+        forms.Form, hg.Lazy, str
+    ] = DEFAULT_FORM_CONTEXTNAME,  # required to derive the widget from a django form field
+    no_wrapper: bool = False,  # wrapper produces less dense layout, from carbon styles
+    no_label: bool = False,
+    show_hidden_initial: bool = False,  # required in special cases to add an initial value
+    #
+    #
+    # --------------------------------------------------------------------------
+    # parameters which are normally not required, when using a django form field
+    # but can be filled in to create form fields independently from form fields or
```
wipascal, line 35:
```suggestion
    # but can be filled in to create form fields independently from django form fields or
```

bread/layout/components/forms/fields.py
```diff
@@ -0,0 +1,189 @@
...
+    # manually overriding values from the form field
+    widgetclass: Optional[
+        BaseWidget
```
wipascal, line 38:
```suggestion
        Type[BaseWidget]
```

bread/forms/forms.py
```diff
@@ -260,19 +259,13 @@ def _formfield_callback_with_request(field, request, model, instance, cache_quer
...
-
 class PreferencesForm(GlobalPreferenceForm):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        self.layout = lambda request: _layout.form.Form.from_fieldnames(
-            hg.C("form"), self.fields
+        self.layout = lambda request: _layout.forms.Form(
+            hg.C("form"),
+            *[_layout.forms.FormField(f) for f in self.fields],
+            _layout.forms.helper.Submit(),
```
wipascal, line 268:
```suggestion
            _layout.forms.helpers.Submit(),
```

wipascal (approved):
Nice work. I think this is a big improvement.
I agree that we need to test this thoroughly...

PR #110, opened 2021-12-16
Added default generic search functionality to BrowseView

dward2nd:
- Improved `get_queryset()` so that it can search as many types of fields as possible.
- Added a default search box (reused and modified from [the Navbar search box in basxconnect](https://github.com/basxsoftwareassociation/basxconnect/blob/main/basxconnect/core/views/person/search_person_view.py) that has a link to the BrowseView with queries.


saemideluxe (approved):
Okay, will merge this version, but continue to work on it.

saemideluxe:
@dward2nd can continue working on:
- Try to use djangoql instead of django Q objects (choose the one that has the easier implementation and works more reliable)
- I think we should actually limit the search to the fields that have been specified inside the tables DataTableColumn, in order to avoid confusion why certain results might be coming up...

PR #113, opened 2022-01-05
make DeleteView model instance available in subclasses

wipascal:
I want to be able to use the instance in a subclass...

### DIFF
--- a/bread/views/delete.py
+++ b/bread/views/delete.py
@@ -39,14 +39,12 @@ class DeleteView(BreadView, PermissionRequiredMixin, RedirectView):
         return [f"{self.model._meta.app_label}.delete_{self.model.__name__.lower()}"]
 
     def get(self, *args, **kwargs):
-        self.instance = get_object_or_404(self.model, pk=self.kwargs.get("pk"))
-        restore = "restore" in self.request.GET
         if self.softdeletefield:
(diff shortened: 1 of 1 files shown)


saemideluxe:
Hmm... a few comments to the delete view in general:
- If we want to make this suitable for a subclass, then setting the instance might be better in the dispatch method I think (https://docs.djangoproject.com/en/dev/ref/class-based-views/base/#django.views.generic.base.View.dispatch)
- It might be nicer if we use a set of function-based view here, for hard-delete, soft-delete and restore. The autopath function would be updated accordingly and I think we would configure "soft-deletablness" via a parameter of the autopath function
- Actually this should be a POST and not a GET (it could also be a DELETE, but that makes only sense in the context of a REST-API) because it modifies the database.

If you do not want to invest too much time here I suggest you just move the instance-loading into the dispatch method, so that the instance would be available in all possible HTTP method handlers.

wipascal:
The nice thing about the class-based views is that it is easy to create instances for different models. If we just have one view function `def hard_delete(request, pk)`, how can this be used for different models? Or are you talking about a function `def generate_hard_delete_view(model)` which returns the actual view function?

wipascal:
never mind, I don't need this anymore for my basxconnect changes

PR #115, opened 2022-01-07
add typechecking of the package `bread/forms` to the CI

wipascal:


bread/layout/components/datatable.py
```diff
@@ -109,7 +109,7 @@ def __init__(
         rowvariable: str = "row",
         spacing: str = "default",
         zebra: bool = False,
-        **kwargs: dict,
+        **kwargs: str,
```
saemideluxe, line 112: Why is this ```str```? The real type of ```kwargs``` will be ```dict```.

saemideluxe:
```python
def foo(**kwargs):
    print(type(kwargs))

foo() # output: <class 'dict'>
```

bread/layout/components/datatable.py
```diff
@@ -132,15 +132,17 @@ def __init__(
         kwargs["_class"] = kwargs.get("_class", "") + " ".join(
             DataTable.tableclasses(spacing, zebra)
         )
-        super().__init__(hg.THEAD(self.head), hg.TBODY(self.iterator), **kwargs)
+        super().__init__(
+            hg.THEAD(self.head), hg.TBODY(self.iterator), lazy_attributes=None, **kwargs
+        )
```
saemideluxe, line 137: This will prevent the parameter ```lazy_attributes``` to be passed to ```DataTable.__init__```.

bread/layout/components/forms/fields.py
```diff
@@ -34,8 +34,8 @@ def generate_widget_element(
     # parameters which are normally not required, when using a django form field
     # but can be filled in to create form fields independently from django form fields or
     # manually overriding values from the form field
-    widgetclass: Optional[
-        Type[BaseWidget]
+    widgetclass: Union[
+        Optional[Type[BaseWidget]], hg.Lazy
```
saemideluxe, line 38: I think this should be ``` Optional[ Union[ Type[ BaseWidget ], hg.Lazy ] ]```.

saemideluxe: The point being that the parameter is optional per se, not only a certain option of the parameter type.

bread/layout/components/forms/fields.py
```diff
@@ -46,8 +46,8 @@ def generate_widget_element(
     errors: Optional[
         List[str]
     ] = None,  # normally be taken from the django form field, will be carbon-ized
-    inputelement_attrs: Optional[
-        dict
+    inputelement_attrs: Union[
+        Optional[dict], hg.Lazy
```
saemideluxe, line 50: Same as above.

bread/layout/components/forms/widgets.py
```diff
@@ -28,9 +29,9 @@ class BaseWidget(hg.DIV):
     django_widget = None
 
     # default attributes which are used to create the input element in a standardized way for many inputs
-    carbon_input_class = ""
-    carbon_input_error_class = ""
-    input_type = None
+    carbon_input_class: Optional[str] = ""
+    carbon_input_error_class: Optional[str] = ""
+    input_type: Optional[str] = None
```
saemideluxe, line 34: According to https://docs.python.org/3/library/typing.html#typing.Optional this should be ```carbon_input_class: str = ""``` because 
> Optional[X] is equivalent to X | None (or Union[X, None]).

bread/layout/components/search.py
```diff
@@ -11,9 +11,9 @@ class SearchBackendConfig(typing.NamedTuple):
     """Describes an endpoint for auto-complete searches"""
 
     url: typing.Any
-    result_selector: str = None
-    result_label_selector: str = None
-    result_value_selector: str = None
+    result_selector: typing.Optional[str] = None
+    result_label_selector: typing.Optional[str] = None
+    result_value_selector: typing.Optional[str] = None
```
saemideluxe, line 16: The typing here is correct, however I think it would make sense to make this ```[str]``` and set the empty string as the default, wouldn'nt it?

bread/views/browse.py
```diff
@@ -67,23 +67,28 @@ class BrowseView(BreadView, LoginRequiredMixin, PermissionListMixin, ListView):
     orderingurlparameter: str = "ordering"
     objectids_urlparameter: str = "_selected"  # see bread/static/js/main.js:submitbulkaction and bread/layout/components/datatable.py
     bulkaction_urlparameter: str = "_bulkaction"
-    items_per_page_options: Tuple[int] = None
+    items_per_page_options: Optional[Tuple[int]] = None
     itemsperpage_urlparameter: str = "itemsperpage"
 
     title: Optional[hg.BaseElement] = None
-    columns: Tuple[Union[str, layout.datatable.DataTableColumn]] = ["__all__"]
+    columns: Tuple[Union[str, layout.datatable.DataTableColumn]] = ("__all__",)
```
saemideluxe, line 74: This should either be ```Tuple[Union[str, layout.datatable.DataTableColumn], ...]``` or maybe better ```Iterable[Union[str, layout.datatable.DataTableColumn]]```. We could restrict it to tuples, but allowing array syntax would make sense.

Okay, to be fully compliant we would need to rewrite ```bread.layout.datatable.DataTable.from_model``` to iterate only once over ```columns``` to allow any iterable (e.g. a generator). Which is possible without problems I think. Or we use ```Sequence[Union[str, layout.datatable.DataTableColumn]]``` instead, but that would feel like an arbitrary limitation.

bread/views/browse.py
```diff
@@ -67,23 +67,28 @@ class BrowseView(BreadView, LoginRequiredMixin, PermissionListMixin, ListView):
...
     # bulkactions: List[(Link, function(request, queryset))]
     # - link.js should be a slug and not a URL
     # - if the function returns a HttpResponse, the response is returned instead of the browse view result
-    bulkactions: Tuple[
-        Link, Callable[[HttpRequest, models.QuerySet], Union[None, HttpResponse]]
+    bulkactions: Union[
+        Tuple[
+            Link, Callable[[HttpRequest, models.QuerySet], Union[None, HttpResponse]]
+        ],
+        Tuple[()],
```
saemideluxe, line 84: The same as above is true for this. And then the appended ```Tuple[()]`` should also no longer be necessary.

bread/layout/components/forms/widgets.py
saemideluxe: (Only for ```carbon_input_class``` and ```carbon_input_error_class```)

bread/layout/components/datatable.py
```diff
@@ -501,8 +505,8 @@ class DataTableColumn(NamedTuple):
     cell: Any
     sortingname: Optional[str] = None
     enable_row_click: bool = True
-    th_attributes: hg.F = None
-    td_attributes: hg.F = None
+    th_attributes: Optional[hg.F] = None
+    td_attributes: Union[hg.F, dict, None] = None
```
saemideluxe, line 509: I suggest this should be written as 

```python
th_attributes: Optional[ Union[ hg.F, dict ] ] = None
td_attributes: Optional[ Union[ hg.F, dict ] ] = None
```

bread/layout/components/datatable.py
```diff
@@ -311,15 +315,15 @@ def from_model(
         model,
         queryset=None,
         # column behaviour
-        columns: List[Union[str, "DataTableColumn"]] = None,
+        columns: Union[Iterable[Union[str, "DataTableColumn"]], Tuple[()]] = (),
```
saemideluxe, line 318: Can we replace this with ```Iterable[Union[str, layout.datatable.DataTableColumn]]```? (see the other comments to the ```columns``` parameter)

bread/layout/components/datatable.py
```diff
@@ -311,15 +315,15 @@ def from_model(
...
         prevent_automatic_sortingnames=False,
         # row behaviour
         rowvariable="row",
-        rowactions: List[Link] = (),
+        rowactions: Union[Iterable[Link], Tuple[()]] = (),
```
saemideluxe, line 322: I think here too we should be able to use just ```Iterable[Link]``` as an empty tuple matches this.

bread/layout/components/datatable.py
```diff
@@ -311,15 +315,15 @@ def from_model(
...
         rowactions_dropdown=False,
         rowclickaction=None,
         # bulkaction behaviour
-        bulkactions: List[Link] = (),
+        bulkactions: Union[Iterable[Link], Tuple[()]] = (),
```
saemideluxe, line 326: I think here too we should be able to use just ```Iterable[Link]``` as an empty tuple matches this.

saemideluxe (changes requested):
Okay, a lot of comments, I think some of the issues might also originate from when I wrote annotations the first time.

Another thing I wonder is: Could it make sense that we implement our own type (see https://docs.python.org/3/library/typing.html#user-defined-generic-types) for the cases where we use ```htmlgenerator.Lazy```? I hope we could write ```Lazy[str]``` or ```Lazy[Any]``` and similar instead of always skipping or using ```Union```.

1. We express the expected type of the Lazy expression
2. No need to always think the "non-lazy" option.

.github/workflows/main.yml
```diff
@@ -65,3 +65,6 @@ jobs:
 
     - name: Test with pytest
       run: ./manage.py test
+
+    - name: Check types
+      run: mypy bread/forms
```
saemideluxe, line 70: Wait, is there a confusion between ```bread/forms``` and ```bread/layout/components/forms``` maybe?

wipascal: Not really. These are the typehints that were necessary for `mypy bread/forms` to succeed...

bread/layout/components/datatable.py
wipascal: At runtime it's a `dict`, but the typehint is certainly not `dict`. If you want I can put `Any` instead of `str`?

wipascal: Yes, it actually has to be `Any`

bread/layout/components/forms/fields.py
wipascal: Yeah I think it's the same thing. `Optional[Type[BaseWidget]]` is the same as `Union[Type[BaseWidget], None]`.
So the typehint could also be `Union[ Type[ BaseWidget ], hg.Lazy, None]`. It's a matter of taste. But I will use your suggestion because it's not really worth a big discussion.

bread/layout/components/datatable.py
wipascal: > Okay, a lot of comments, I think some of the issues might also originate from when I wrote annotations the first time.
> 
> Another thing I wonder is: Could it make sense that we implement our own type (see https://docs.python.org/3/library/typing.html#user-defined-generic-types) for the cases where we use `htmlgenerator.Lazy`? I hope we could write `Lazy[str]` or `Lazy[Any]` and similar instead of always skipping or using `Union`.
> 
>     1. We express the expected type of the Lazy expression
> 
>     2. No need to always think the "non-lazy" option.

I like the idea of introducing "shortcuts" for the types. They can get very verbose...
But `Lazy[str]` to me just looks like a Lazy that evaluates to a string and not like `Union[Lazy, str]`. I will need to think about this a bit more. Maybe we can introduce an alias for 'Union[Lazy, str, hg.BaseElement]`. But what would we call that? (I guess you would hate `Renderable` ;) )

wipascal:
> Okay, a lot of comments, I think some of the issues might also originate from when I wrote annotations the first time.
> 
> Another thing I wonder is: Could it make sense that we implement our own type (see https://docs.python.org/3/library/typing.html#user-defined-generic-types) for the cases where we use `htmlgenerator.Lazy`? I hope we could write `Lazy[str]` or `Lazy[Any]` and similar instead of always skipping or using `Union`.
> 
>     1. We express the expected type of the Lazy expression
> 
>     2. No need to always think the "non-lazy" option.

I like the idea of introducing "shortcuts" for the types. They can get very verbose...

But `Lazy[str]` to me just looks like a Lazy that evaluates to a string and not like `Union[Lazy, str]`. For me we would need to call it `MaybeLazy[str]` or something like that. 

(Actually I like the idea of using `Lazy[T]` for a Lazy that evaluates to T, but that's a different topic altogether.)

bread/layout/components/forms/fields.py
saemideluxe: Agreed, it is the same. But thanks for changing.

bread/views/browse.py
saemideluxe: Can you make this an iterable as well?

bread/views/browse.py
```diff
@@ -67,23 +67,28 @@ class BrowseView(BreadView, LoginRequiredMixin, PermissionListMixin, ListView):
     orderingurlparameter: str = "ordering"
     objectids_urlparameter: str = "_selected"  # see bread/static/js/main.js:submitbulkaction and bread/layout/components/datatable.py
     bulkaction_urlparameter: str = "_bulkaction"
-    items_per_page_options: Tuple[int] = None
+    items_per_page_options: Optional[Tuple[int]] = None
```
saemideluxe, line 70: And this should be ```Optional[Iterable[int]]``` too I think. It is used to declare which options for the "N-items per page" dropdown will show.

saemideluxe:
> > Okay, a lot of comments, I think some of the issues might also originate from when I wrote annotations the first time.
> > Another thing I wonder is: Could it make sense that we implement our own type (see https://docs.python.org/3/library/typing.html#user-defined-generic-types) for the cases where we use `htmlgenerator.Lazy`? I hope we could write `Lazy[str]` or `Lazy[Any]` and similar instead of always skipping or using `Union`.
> > ```
> > 1. We express the expected type of the Lazy expression
> > 
> > 2. No need to always think the "non-lazy" option.
> > ```
> 
> I like the idea of introducing "shortcuts" for the types. They can get very verbose...
> 
> But `Lazy[str]` to me just looks like a Lazy that evaluates to a string and not like `Union[Lazy, str]`. For me we would need to call it `MaybeLazy[str]` or something like that.
> 
> (Actually I like the idea of using `Lazy[T]` for a Lazy that evaluates to T, but that's a different topic altogether.)

Right, ```MaybeLazy[T]``` and ```Lazy[T]``` would be interesting, if somebody needs something recreational to do, that could be added in htmlgenerator ;)

saemideluxe (approved):

PR #116, opened 2022-01-07
Add tooltip components

dward2nd:
I have completed the component classes for the Carbon-based tooltips. Classes, IDs, and attributes are based on the [given example](https://the-carbon-components.netlify.app/?nav=tooltip).


bread/layout/components/tooltip.py
```diff
@@ -0,0 +1,282 @@
...
+    """Bring the class in 'source' inte 'dest'"""
+    if "_class" in source:
+        if source["_class"]:
+            dest["_class"] += source["_class"]
+        del source["_class"]
+
+    # for other attributes, if there exist on both dicts,
+    # just replace them.
+    for attr in dest.keys() & source.keys():
+        dest[attr] = source[attr]
```
saemideluxe, line 32: Could you replace this function with ```htmlgenerator.merge_html_attrs```? We added that some time ago to avoid having to rewrite the same function for every component.

bread/layout/admin.py
```diff
@@ -165,6 +166,7 @@ class ReindexForm(forms.Form):
             ),
             style="margin-bottom: 1rem;",
         ),
+        tooltip.InteractiveTooltip("Test Label", "Test Body"),
         reindex_btn,
         hg.If(
             logmsg,
```
saemideluxe, line 172: Can we remove the changes in this file? If we want to have something for preview on the UI I recommend you add it on the view ```bread.views.admin.widgetpreview```.

saemideluxe (changes requested):
Well done, especially adding the documentation from carbon. I think this is something we should plan to in the future do for all components.

bread/layout/admin.py
dward2nd: I forgot to remove that. Sorry.

bread/layout/components/tooltip.py
dward2nd: OK. Done.

bread/layout/components/tooltip.py
```diff
@@ -0,0 +1,269 @@
...
+from bread.layout.components.icon import Icon
+from bread.utils import Link
+
+
+def _get_icon(icon):
+    """Return a proper Icon instance"""
+    if isinstance(icon, str):
+        icon = Icon(icon, size="16")
+    else:
+        icon.attributes["_class"] = icon.attributes.get("_class", "") + " bx--btn__icon"
```
saemideluxe, line 15: Sorry, I forgot to mention this one before: htmlgenerator.lazy objects do currently not support the plus operator. Therefore we have to write it like:

```hg.BaseElement(icon.attributes.get("_class", ""), " bx--btn__icon")```

We will add support for expression for htmlgenerator.Lazy in the future, but not yet, so we need to wrap this with a BaseElement.

Comment (wipascal) [inline bread/layout/components/tooltip.py:15]: Can you explain me why this is needed? If the parameter is already of type `Icon`, surely the class `bx--btn__icon` will already be there. if the parameter is `str`, the instanciation of Icon will set the class `bx--btn__icon` as well. So in what scenario is this needed?

Comment (dward2nd) [reply bread/layout/components/tooltip.py:15]: I might have seen this one in the documentation, so I did that. But I think I'd rather remove it as you suggested. Thanks for the suggestion.

Comment (dward2nd) [reply bread/layout/components/tooltip.py:15]: Thank you for the information. Though I later discovered there's no need to add the class for that one.

Comment (wipascal) [reply bread/layout/components/tooltip.py]: Actually I am not sure Icon add the class `bx--btn__icon` automatically. But then you would need to add it to the instance you create on line 13 as well.

Comment (saemideluxe) [reply bread/layout/components/tooltip.py]: Icons have no class by default. The ```"bx--btn__icon"``` is normally used for clickable icon I think.

File: bread/layout/components/tooltip.py [hunk 4]
```diff
@@ -0,0 +1,268 @@
...
+        """
+        tooltip_attributes = {
+            "_class": "bx--tooltip--definition bx--tooltip--a11y ",
+            "data_tooltip_definition": True,
+        }
+
+        tooltip_attributes = hg.merge_html_attrs(tooltip_attributes, attributes)
+        asst_txt_id = hg.html_id(self, "bx--tooltip--definition-id")
+
+        super().__init__(
```
wipascal, line 64: you can save two lines of code here by extending `DIV` directly insteady of `BaseElement`, but it's no big deal

bread/layout/components/tooltip.py
```diff
@@ -0,0 +1,268 @@
...
+class InteractiveTooltip(hg.BaseElement):
+    """Interactive tooltip should be used if there are actions a user can take in
+    the tooltip (e.g. a link or a button). For more regular use case, e.g. giving
+    the user more text information about something, use definition tooltip or icon
+    tooltip.
+
+    Reference: https://the-carbon-components.netlify.app/?nav=tooltip
+    """
+
+    base_class = "bx--tooltip"
```
wipascal, line 155: I would move these into the constructor because having them here suggests that you are using them in a subclass or sharing them between multiple methods, which you are not doing currently.

bread/layout/components/tooltip.py
```diff
@@ -0,0 +1,268 @@
...
+    suffix_label = "label"
```
wipascal, line 156: It this variable only used in one place? If so, I would inline it

bread/layout/components/tooltip.py
```diff
@@ -0,0 +1,268 @@
...
+            hg.BaseElement(
+                hg.If(bool(link), hg.A(link.label, href=link.href, _class="bx--link")),
+                hg.If(bool(button), button),
+            )
+            if link or button
+            else None
+        )
+
+        base_id = hg.html_id(self, self.base_class + "-id")
+        label_id = "-".join([base_id, self.suffix_label])
```
wipascal, line 200: I would use an f-string instead of a join here, since you are only joining two strings

bread/layout/components/tooltip.py
dward2nd: In some cases, like `InteractiveTooltip`, it doesn't come as a whole element of `<div>`, but it's separated into two parts, one for a label, and another for the tooltip itself.

bread/layout/components/tooltip.py
dward2nd: Hmm... I think so. Thank you.

bread/layout/components/tooltip.py
dward2nd: Not really. They're reused at several parts within the code. Some may have different variation, like `bx--tooltip__label`. So I declare those variables to avoid typos from retyping.

bread/layout/components/tooltip.py
dward2nd: Thank you for the suggestion.

bread/layout/components/tooltip.py
dward2nd: As I took a look again in the documentation, there's no `bx--btn__icon` used in tooltips. So that would be my misunderstanding.

bread/layout/components/tooltip.py
saemideluxe: @dward2nd What K. Pascal means is: Is this value only ever used inside the ```__init__``` method?
If yes, then you could limit the scope of the variable to be only inside ```__init__``` instead of the whole class.

saemideluxe (commented):
For all the ```__init__``` parameters: label, description, body and heading:
Can you change type annotation to be Any? The problem is that we want to accept ```hg.BaseElement``` in order to render something, but also str or anything else than can be converted to a string.

saemideluxe (approved):
Great, thanks.

Issue #35, opened 2021-05-25
Add 5XX and 4XX status pages for production


saemideluxe:
We should have a nice 4XX page for production.

Also the 5XX page is necessary IMHO. In general it is not recommended to show too much error information to the user on a production system due to security issues (revealing paths, versions and code). However, I think that a sensible 5XX page with the stacktrace would be a better user experience. The standard plain 500 page leaves users very confused. Showing the stack-trace end the exception will make it more clear that something is not right and we are more likely to get a screenshot with usefull debug information on first contact. But in any case I would limit this to authenticated users.

Further information about this can be found in the Django docs:
https://docs.djangoproject.com/en/dev/topics/http/views/#customizing-error-views

saemideluxe:
Edit: Do not show the stack-trace, only the exception message.

dward2nd:
Done #117

PR #117, opened 2022-01-09
Default status pages for production

dward2nd:
Due to this issue #35, I have made views for status codes 400, 403, 404, and 500 (The default views that can be overrided from Django).

In order to test, temporarily set `DEBUG = False` in the setting first. And run the server using `./manage.py runserver --insecure`. This is because Django won't manage static files when debug mode turned off.

I've tested the views for 404 and 500. However, I still have no idea how to test 403 or 400 for now. But I hope it would give the similar result as well.


bread/layout/error.py
```diff
@@ -0,0 +1,59 @@
...
+
+@aslayout
+def error_layout(
+    request,
+    status_code: int,
+    status_title: str,
+    description: Union[str, hg.BaseElement],
+    exception: BaseException = None,
+    exception_detail: str = None,
+    redirect_link: Link = None,
```
saemideluxe, line 17: To simplify the signature a bit I recommend you remove ```redirect_link``` (as it is not used) and remove the ```exception``` parameter and instead always pass ```exception_details=exception.message``` when calling ```error_layout```

saemideluxe: (This should also allow you to simplify the ```if``` statments below a bit, which makes it easier to read and understand.

bread/views/error.py
```diff
@@ -0,0 +1,78 @@
...
+
+import htmlgenerator as hg
+from django.http import HttpResponse, HttpResponseNotFound
+from django.utils.translation import gettext_lazy as _
+
+from bread.layout import error as layouts
+
+
+def view400(request, exception):
+    return HttpResponse(
```
saemideluxe, line 11: When you call a function that has a ```@aslayout``` decorator, the response should already be an HttpResponse, so no need to wrap it here. Same should be true for the other views.
In order to set the status code for the response I think you should be able to do something like this:

```python
resp = layout.error_layout(.....)
resp.status = 400
return resp
```

dward2nd: I didn't know about that. Thank you.

bread/layout/error.py
dward2nd: OK. Done.

bread/views/error.py
```diff
@@ -0,0 +1,78 @@
...
+                    _(
+                        "There's a problem occurred within the server related to path %s."
+                        "<br>If the problem persists, please contact admin."
+                    )
+                    % f"<strong>{request.path}</strong>"
+                )
+            ),
+            exception_detail=f"{str(exec_info[0].__name__)}: {exec_info[1]}",
+        )
+    )
```
saemideluxe, line 74: Are view 404 and view500 reponses not missing the status code in the response?

saemideluxe (changes requested):
Okay, looks good. Just a few small changes.

bread/views/error.py
dward2nd: According to Django, the view for status code 500 doesn't receive `exception` as a parameter (only `request` is required). So that's why I have to retrieve exception information from `sys.exc_info`.

For 404, it does provide an exception. However, when I was testing the bug, it didn't provide any detail (both `exception.message` and the string representation of `exception`). It only came with a list of all available `URLConf`, which is rather a debug detail than exception one.

bread/layout/error.py
```diff
@@ -0,0 +1,40 @@
...
+                        "border: 1px solid grey;"
+                        "padding: 1rem;"
+                        "font-family: monospace;"
+                        "margin-bottom: 1rem;"
+                    ),
+                ),
+            ),
+        ),
+        Button.fromlink(Link(label="Back to homepage", href="/")),
+    )
```
saemideluxe, line 40: One more thing: Could you please move this function into the errors view? Two reasons:
1. It is actually a view and not a layout (the ```@aslayout``` decorator will convert the function into a view)
2. I think you are trying to following the principle of "separation of concerns", right? But you also need to keep in mind the principle of "locality of behaviour" (see https://htmx.org/essays/locality-of-behaviour/ for more about it). In this case, having the layout close to the view makes maintenance easier as the layout will never need to be referenced from other places in the codebase.

bread/views/error.py
saemideluxe: Yes, that is okay.

bread/layout/error.py
```diff
@@ -0,0 +1,40 @@
...
+                    style=(
+                        "border: 1px solid grey;"
+                        "padding: 1rem;"
+                        "font-family: monospace;"
+                        "margin-bottom: 1rem;"
+                    ),
+                ),
+            ),
+        ),
+        Button.fromlink(Link(label="Back to homepage", href="/")),
```
saemideluxe, line 39: Please make the "Back to homepage" string translatable. Also, href should not simply be the root. The reason for this is, that sometimes we host an application on a sub-path of a domain. In such cases we only want the web-application root, not the root of the domain. You can see an example of this here: https://github.com/basxsoftwareassociation/bread/blob/main/bread/layout/components/shell_header.py#L83

```href=hg.F(lambda c: c["request"].META["SCRIPT_NAME"] or "/")```

bread/layout/error.py
dward2nd: OK. I'll combine them. Though I didn't know it's called "separation of concerns", but yes, it would make files cleaner for me.

bread/layout/error.py
dward2nd: OK. Done.

saemideluxe (approved):
Okay, should be fine now.

PR #118, opened 2022-01-10
Feature/defaultsearchforbrowseviews restructured

saemideluxe:
I wanted to simplify a few things with the default searchbar, therefore I branched from you current PR.


bread/layout/components/datatable.py
```diff
@@ -257,15 +267,28 @@ def with_toolbar(
...
+                                        != search_urlparameter
+                                    ),
+                                    hg.INPUT(
+                                        type="hidden",
+                                        name=hg.C("queryitem")[0],
+                                        value=hg.C("queryitem")[1],
+                                    ),
+                                ),
+                            ),
+                            method="GET",
```
saemideluxe, line 287: This is to allow the search to work without javascript. However, clearing the searchbar through the button still requires javascript.

bread/layout/components/search.py
```diff
@@ -105,9 +105,12 @@ def _loading_indicator(resultcontainerid):
...
 
 
-def _close_button(resultcontainerid):
+def _close_button(resultcontainerid, widgetattributes):
     kwargs = {
-        "_class": "bx--search-close bx--search-close--hidden",
+        "_class": hg.BaseElement(
+            "bx--search-close",
+            hg.If(widgetattributes.get("value"), None, " bx--search-close--hidden"),
+        ),
```
saemideluxe, line 113: This should be here anyway, was missing. In order to show the close-button always if there is some value inside the search bar.

bread/utils/queryset_from_fields.py
```diff
@@ -78,6 +64,8 @@ def get_field_queryset(fields, queries, prefix="", **kwargs):
                 foreign_fields = foreignkey_field._meta.fields
 
             new_prefix = prefix + "__".join([foreignkey_field.name, ""])
-            qs |= get_field_queryset(foreign_fields, queries, new_prefix, **kwargs)
+            qs |= get_field_queryset(
+                foreign_fields, searchquery, new_prefix, follow_relationships - 1
+            )
 
     return qs
```
saemideluxe, line 71: This implementation seems to work pretty good. I adjusted the recursion a bit (so we could also go deeper if necessary) and I wanted the query to be a single string, not a list. More complicated queries should be done qith djangoql.

bread/views/browse.py
```diff
@@ -192,15 +183,7 @@ def get(self, *args, **kwargs):
...
-                if self.request.GET["reset"] == "2":
-                    # only reset search value
-                    session_str = self.request.session[self.viewstate_sessionkey].split(
-                        "&"
-                    )
-                    session_str = "&".join((x for x in session_str if "q=" not in x))
-                    self.request.session[self.viewstate_sessionkey] = session_str
-                else:
-                    del self.request.session[self.viewstate_sessionkey]
+                del self.request.session[self.viewstate_sessionkey]
```
saemideluxe, line 186: I wanted the searchquery to use the same reset-mechanism as any other filters, therefore removed some of the logic.

bread/views/browse.py
```diff
@@ -248,15 +231,24 @@ def get_paginate_by(self, queryset):
...
+                        _("Bad filter string '%s': '%s'") % (searchquery, e),
+                    )
+
+            else:
+                qs = self.model.objects.filter(
+                    queryset_from_fields.get_field_queryset(
+                        [*self.model._meta.fields, *self.model._meta.many_to_many],
+                        searchquery,
+                    )
+                )
```
saemideluxe, line 251: Added this: When starting with the = sign we can use djangoql for the search string.

bread/layout/components/datatable.py
```diff
@@ -227,6 +226,19 @@ def with_toolbar(
                 ),
             )
 
+        searchbar = Search(
```
wipascal, line 229: I find it a bit confusing that we do all this work even if search_urlparameter is `None` and the 'searchbar' variable goes unused

bread/layout/components/datatable.py
```diff
@@ -227,6 +226,19 @@ def with_toolbar(
...
+            widgetattributes={
+                "autofocus": True,
+                "name": search_urlparameter,
+                "value": hg.F(
+                    lambda c: html.escape(c["request"].GET.get(search_urlparameter, ""))
+                ),
+                "onfocus": "this.setSelectionRange(this.value.length, this.value.length);",
+            }
+        )
+        searchbar[0][3].attributes[
```
wipascal, line 239: I think it would be appropriate to have a field `searchbar.close_button` instead of these magic indices

bread/layout/components/search.py
wipascal: how does `None` get rendered?

bread/layout/components/datatable.py
wipascal: I would actually prefer adding close_button_attributes to the constructor parameters of Search (I admit to be a fan of immutability).

bread/layout/components/datatable.py
```diff
@@ -257,15 +269,28 @@ def with_toolbar(
...
+                                        value=hg.C("queryitem")[1],
+                                    ),
+                                ),
+                            ),
+                            method="GET",
                         ),
                         _class="bx--toolbar-search-container-persistent",
                     )
-                    if search_backend
+                    if search_urlparameter is not None
```
wipascal, line 293: maybe replace this with `hg.If` (the ternary operator in Python has earned many awards for being one of the ugliest features in the Language)

bread/layout/components/datatable.py
```diff
@@ -139,18 +140,17 @@ def with_toolbar(
         title: Any,
         helper_text: Any = None,
         primary_button: Optional[Button] = None,
-        search_backend: Optional[SearchBackendConfig] = None,
         bulkactions: List[Link] = (),
         pagination_config: Optional[PaginationConfig] = None,
         checkbox_for_bulkaction_name="_selected",
+        search_urlparameter=None,
```
wipascal, line 146: maybe let's add typehints for all parameters (since most of them already have it)

bread/views/browse.py
dward2nd: If DjangoQL tends to work better in most cases (like complicated search), why don't we replace the old one with entirely DjangoQL then?

bread/layout/components/datatable.py
dward2nd: This is even harder to understand for new developers to continue.

bread/layout/components/datatable.py
dward2nd: This is great. It looks a lot simpler than the JS one.

bread/layout/components/datatable.py
```diff
@@ -139,18 +140,17 @@ def with_toolbar(
         title: Any,
         helper_text: Any = None,
         primary_button: Optional[Button] = None,
-        search_backend: Optional[SearchBackendConfig] = None,
```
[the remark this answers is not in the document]
saemideluxe: Right, good thinking. But do not worry about deleting too much code. With git we can always go back or merge something from an older revision.

bread/layout/components/datatable.py
saemideluxe: Right, done.

bread/layout/components/datatable.py
saemideluxe: I tend to use ```hg.If ```only if the condition could be lazy-evaluated. But I do not really see a sensible use case where ```search_urlparameter``` would need to be lazy. Therefore relying on Pythons conditional operator makes more sense here I think.

Right, I read about criticism a few times too .But I wonder if there is more to it than personal opinion. People who are used to the C-style ternary operator ```? :``` are likely to find this way of ordering a conditional operator confusing, but I find it actually quite readable.

bread/layout/components/search.py
saemideluxe: ```htmlgenerator``` will ignore ```None``` elements when rendering. This is a little "trick" because we have no lazy ```not``` operator, so we simply switch the if and the else parts.

bread/views/browse.py
saemideluxe: Hmm... That might actually be a good idea. I already looked at it briefly in the past, but because we still need to make distinctions between different field types I thought it would not add that much value (e.g. we have to generate ```'id = 3'``` and not ```'id = "3"'```).
However, it might still make sense to simplify ```get_field_queryset``` maybe?

bread/layout/components/datatable.py
saemideluxe: Okay, so I refactored this quite a bit, this should look better.

Regarding the close button: I prefere  a ```close_button``` attribute over an additional parameter ```close_button_attributes```, as it is an easier to use API. In general I would not try to enforce immutability on components, mainy because htmlgenerator and Python are both not designed for immutability. It allows for an easier implementation and API in many cases.

bread/layout/components/datatable.py
saemideluxe: RIght, I just discovered that this is a more elegant way to do it. I think there are other places in the code where we can change it to use hidden inputs to preserve the URL parameters.

bread/layout/components/datatable.py
saemideluxe: Right, refactored that.

wipascal (approved):
Thanks for the improvements!

bread/layout/components/datatable.py
wipascal: I guess the problem originates from the order being different in the if-else statements and in the if-else expressions. That's what makes it unintuitive for many people.

bread/layout/components/datatable.py
wipascal: I think every small step towards immutability helps. But I think we can just agree that we disagree ;)

saemideluxe: :D Okay, let's leave it at that.

PR #119, opened 2022-01-12
Feature/mypy all of bread

wipascal:


bread/views/edit.py
```diff
@@ -24,8 +25,8 @@ class EditView(
     """TODO: documentation"""
 
     accept_global_perms = True
-    fields = None
-    urlparams = (("pk", int),)
+    fields: Optional[List[str]] = None
```
saemideluxe, line 28: Fields can also be hg elements, this is helpfull to customize single fields without having to overwrite the whole layout method.
```fields: Optional[List[Union[hg.BaseElement, str]]] = None```

File: bread/views/edit.py [hunk 2]
```diff
@@ -24,8 +25,8 @@ class EditView(
...
+    urlparams: Union[Tuple[Tuple[str, Type]], Tuple[()]] = (("pk", int),)
```
saemideluxe, line 29: I think this should be 

```urlparams: Iterable[Tuple[str, Type]] = (("pk", int),)```

File: bread/views/userprofile.py [hunk 3]
```diff
@@ -292,7 +292,7 @@ def get_object(self):
 class EditPermissionsView(EditView):
     model = get_user_model()
     fields = ["is_active", "is_superuser", "is_staff"]
-    urlparams = {}
+    urlparams = ()
```
saemideluxe, line 295: I think we should add ```urlparams: Iterable[Tuple[str, Type]] = ()``` in ```bread.views.util.BreadView``` instead. This feature is used just so the URL paths can be generated automatically for class-based views.

saemideluxe (commented):
Are these really *all* the changes required for mypy in all of bread?
The mypy command is still only running for bread/forms.

wipascal:
> Are these really _all_ the changes required for mypy in all of bread? The mypy command is still only running for bread/forms.

It seems so, yes. Now only python 3.7 seems to be a problem for the mypy execution on the whole of bread.

saemideluxe:
Do you plan to do something about the 3.7 type error or just allow it fail? In case you want to make it work, it should be trivial to download Python 3.7 into a temporary directory and set up and activate a 2nd venv which runs mypy with Python 3.7, I did this already a few times in the past.

saemideluxe (approved):
Okay, nice!

PR #122, opened 2022-01-16
Missing Tab content fix

dward2nd:
According to [this issue in basxconnect's repo](https://github.com/basxsoftwareassociation/basxconnect/issues/218), this is because `Tabs` depend too much on the cookie value. So, the solution is to check the cookie value before it's been used.


saemideluxe (commented):
The code is basically okay. However, I think in this case using a samll piece of javascript could be a bit more robust. I am thinking something like:

```python
"onload"="if( $('.bx--tabs__nav-item--selected', this) == null ) (new CarbonComponents.Tab(this)).setActive($('.bx--tabs__nav-item', this))"
```
on the Tabs.labelcontainer element. My idea is that this captures every possible case where there is visually no tab selected. Normally I prefer doing it server-side, as you have done here. But in this case the javascript logic seems much simpler than the serverside handling of the cookie value.

What do you think?

dward2nd:
I think neither is better in terms of user experience or performance. So I'll leave it to your decision then.

saemideluxe:
Okay, pushed the JS fix. Can you test this too in your environment?

saemideluxe:
For you information: the "onload" attribute is not javascript/DOM standard for most HTML elements. However, because it is often usefull to do some initialization when writing standalone interface components we do call the onload attribute artifically with bread.

dward2nd:
Works normally as well as the cookie one.

saemideluxe:
Ok, thanks.

saemideluxe (approved):

PR #124, opened 2022-01-18
Feature/test tooltips

wipascal:


dward2nd (approved):
I think the widget preview should contain more usage examples of the tooltip, especially possible directions and alignments it can be.

saemideluxe (changes requested):
Could add the mentioned change in this PR as well? Regarding the replacement of BUTTON with DIV.

saemideluxe (approved):
Cool, thanks!

saemideluxe:
Ah, isort failing btw.

PR #126, opened 2022-01-20
first (very basic) implementation of document templates

wipascal:
This is mainly to find out if I am going in the right direction.

Todo's for future PRs:
- add a "download zip" button which allows for downloading all the generated documents at once
- improve UI (once we get more clarity on the desired functionality)

Maybe (?):
- allow to download documents as PDFs (instead of .docx) 
- limit the number of documents that is generated each time the button is saved. 
    - Maybe generate the documents only if a "generate documents from template" checkbox is ticked
    - or introduce a separate button "generate documents from template"
    - or generate only a couple of samples, unless a checkbox "generate documents for all objects matching filter" is ticked

@saemideluxe What do you think?


bread/contrib/document_templates/urls.py
```diff
@@ -0,0 +1,32 @@
+from bread import menu, views
+from bread.contrib.document_templates.models import DocumentTemplate
+from bread.contrib.document_templates.views import DocumentTemplateEditView
+from bread.utils import Link, default_model_paths
+from build.lib.bread.utils import urls
```
saemideluxe, line 5: Strange import ;)

wipascal: I agree :)
It's fixed.

wipascal: I agree :)
It's fixed.

bread/contrib/document_templates/views.py
```diff
@@ -0,0 +1,70 @@
...
+            ),
+            hg.H4("Rendered documents:"),
+            hg.Iterator(
+                hg.F(lambda c: c["object"].documents.all()),
+                "document",
+                hg.DIV(
+                    hg.A(
+                        hg.C("document").file,
+                        href=hg.format(
+                            "{}/{}", settings.MEDIA_URL, hg.C("document").file
```
saemideluxe, line 35: What about ```hg.C("document").url```?

bread/contrib/document_templates/views.py
```diff
@@ -0,0 +1,70 @@
...
+    )
+
+    buf = io.BytesIO()
+    template.save(buf)
+    buf.seek(0)
+
+    document_name = (
+        f"{os.path.basename(template_path).split('.')[0]}_object{object.id}.doc"
+    )
+    response = HttpResponse(buf.read(), content_type="application/vnd.ms-excel")
```
saemideluxe, line 68: Is the output document a ```*.doc``` or a ```*.docx```? Also the mime type is wrong here I guess.

bread/contrib/document_templates/views.py
```diff
@@ -0,0 +1,70 @@
...
+    def get_success_url(self):
+        return self.request.get_full_path()
+
+
+def generate_document_view(request, template_id: int, object_id: int):
+    document_template = DocumentTemplate.objects.get(id=template_id)
+    object = document_template.model.model_class().objects.get(id=object_id)
+    template_path = document_template.file.path
+    template = DocxTemplate(template_path)
+    variables = template.get_undeclared_template_variables()
```
saemideluxe, line 52: Nitpick: I would avoid some of the single-use variables, but it might be a matter of taste ;)

bread/contrib/document_templates/models.py
```diff
@@ -0,0 +1,25 @@
...
+from bread.contrib.reports.fields.queryfield import QuerysetField
+
+
+class DocumentTemplate(models.Model):
+    name = models.CharField(_("Name"), max_length=255)
+    model = models.ForeignKey(
+        ContentType,
+        on_delete=models.PROTECT,
+    )
+    filter = QuerysetField(_("Filter"), modelfieldname="model")
```
saemideluxe, line 14: What is the exact use of model and filter here? ```filter``` I did not see used. ```model``` I understand, but even there I do not see a real advantage, as the only use seems to be determining the model inside the generation function (which could also be passed via URL, as the caller knows the correct model.

wipascal: Yup, filter is only legacy code.
Would you prefer encoding the model into the URL? I thought this is tricky because we would need a system to encode and decode the name of the model into a URL.
Would an url like that be fine? `document-template/<id>/app/<appname>/model/<modelname>/object/<object_id>`
then I think I could could do `django.apps.apps.get_model(<appname>,<modelname>)`

saemideluxe: I think if we would do something like checks or field-suggestions for a template, then specifying the model as a field would be useful. But as for now there seem no use, so yes, in the URL would be preferable.

Regarding encoding I suggest we use the same that is used by ForeignKey fields: "appname.modelnname"
This works also with apps: ```django.apps.apps.get_model("mayapp.mymodel")``` (which is case-insensitive btw).

bread/contrib/document_templates/views.py
```diff
@@ -0,0 +1,70 @@
...
+            for variable in variables
+        }
+    )
+
+    buf = io.BytesIO()
+    template.save(buf)
+    buf.seek(0)
+
+    document_name = (
+        f"{os.path.basename(template_path).split('.')[0]}_object{object.id}.doc"
```
saemideluxe, line 63: Is there a specific reason for using ```*.doc```? The API seems to suggest that this is a ```*.docx``` file.

bread/contrib/document_templates/models.py
```diff
@@ -0,0 +1,22 @@
...
+
+    def __str__(self):
+        return self.name
+
+
+class Document(models.Model):
+    template = models.ForeignKey(
+        DocumentTemplate, on_delete=models.CASCADE, related_name="documents"
+    )
+    file = models.FileField(upload_to="documents/")
```
saemideluxe, line 22: This is no longer used as well, right?
I think if we need to have a persistent file, then we also could reuse the ```ponds.models.utils.Document``` one.

saemideluxe (commented):
Okay, looks good to me. One thing more that just jumped to my mind: We should add a method on ```DocumentTemplate``` which accepts a Model instance and returns the correct URL. I suggest we use ```bread.utils.links.ModelHref``` to allow for a lazy object (but not for the template object itself, that should in most cases be concrete I think).

wipascal:
> Okay, looks good to me. One thing more that just jumped to my mind: We should add a method on `DocumentTemplate` which accepts a Model instance and returns the correct URL. I suggest we use `bread.utils.links.ModelHref` to allow for a lazy object (but not for the template object itself, that should in most cases be concrete I think).

Hmm I don't understand what you mean. Could you sketch how you would want to use such a method?

saemideluxe:
> Hmm I don't understand what you mean. Could you sketch how you would want to use such a method?

Something like (names are just very long to make intent clear):

```python
# assuming e.g. inside bread.views.ReadView or similar
employment_contract_template = DocumentTemplate.objects.get(name="Employment Contract")

return hg.DIV(
    hg.A("Download Employment Contract", href=employment_contract_template.url_for_filled_out_with(hg.C("object")))
)

```

Does that make sense? I expect that is how we use if most often.

saemideluxe (approved):
Ok, looks good.

saemideluxe:
Ah, I think the new black version (now out of beta for the first time btw) has some changes.

PR #127, opened 2022-01-25
DjangoQL Search Implementation in BrowseView

dward2nd:
This is the continuous work from the previous pull request #110 suggested by @saemideluxe.

To enter the advanced search mode, type `=` at the beginning of the search text box.

Screen Recording 2022-01-25 at 午後04 02 20

This will bring the DjangoQL completion as well. Users can begin to type in using DjangoQL syntax instead of finding the substring like we used to do it before.

### Known Issues

- DjangoQL completion cannot identify the children model's field of the BrowseView. For instance, in basxconnect, the Person BrowseView is the BrowseView for the Person model, which NaturalPerson derived from. There are `first_name` and `last_name` fields in NaturalPerson but DjangoQL cannot recognize them. This can be problematic when you need to search something more specific.

- The normal mode still uses the `Q()` solution instead of the DjangoQL's `apply_search()`, which may be continued soon.


bread/querysetfield/__init__.py
```diff
@@ -58,7 +58,7 @@ def deconstruct(self):
         name, path, args, kwargs = super().deconstruct()
         return name, path, args, {"modelfieldname": self.modelfieldname, **kwargs}
 
-    def contribute_to_class(self, cls, name):
+    def contribute_to_class(self, cls, name, **kwargs):
         super().contribute_to_class(cls, name)
```
saemideluxe, line 62: If we take kwargs we should also pass them to the call to super. Otherwise they would get silently lost.

bread/layout/components/datatable.py
```diff
@@ -566,10 +569,55 @@ def searchbar(search_urlparameter: str):
...
+    #                         value=hg.C("urlvalue"),
+    #                     ),
+    #                 ),
+    #             ),
+    #         ),
+    #         method="GET",
+    #     ),
+    #     _class="bx--toolbar-search-container-persistent",
+    # )
+
```
saemideluxe, line 620: Please remove commented out code.

bread/layout/components/datatable.py
```diff
@@ -566,10 +569,55 @@ def searchbar(search_urlparameter: str):
...
             ),
             method="GET",
+            name="browseviewsearchform",
         ),
+        hg.LINK(
+            rel="stylesheet",
+            type="text/css",
+            href=staticfiles_storage.url("djangoql/css/completion.css"),
+        ),
+        hg.SCRIPT(src=staticfiles_storage.url("djangoql/js/completion.js")),
```
saemideluxe, line 579: If these asset files are only used within the BrowseViewSearch then we should move them over there.

bread/layout/components/forms/queryfield.py
```diff
@@ -0,0 +1,201 @@
...
+import htmlgenerator as hg
+from django.urls import reverse
+from django.utils.translation import gettext_lazy as _
+from djangoql.schema import DjangoQLSchema
+from djangoql.serializers import DjangoQLSchemaSerializer
+
+from bread.layout.components.icon import Icon
+
+
+class BrowseViewSearch(hg.DIV):
```
saemideluxe, line 12: What is the reason that we cannot reuse and adjust the existing Search element? It feels like we add a lot of duplication here.

bread/layout/components/forms/queryfield.py
```diff
@@ -0,0 +1,201 @@
...
+            "search",
+            size=16,
+            _class="bx--search-magnifier",
+            id="searchinputicon_normal",
+            aria_hidden="true",
+        )
+        advanced_inputid = "advanced_search__" + hg.html_id(self)
+        advanced_input = hg.If(
+            model,
+            hg.TEXTAREA(
```
saemideluxe, line 58: This can also be a normal input element, no need for it to be a textarea.

saemideluxe (commented):
I think I want a few more changes. My biggest question would be: Is there a specific reason, the we cannot use a single input element to handle both cases?

bread/layout/components/forms/queryfield.py
```diff
@@ -0,0 +1,201 @@
...
+                _close_button(None, {"value": defaultvalue}),
+                **kwargs,
+            ),
+            hg.If(
+                model,
+                hg.SCRIPT(
+                    hg.format(
+                        """
+                document.addEventListener("DOMContentLoaded", () => {{
+                    const inputContainer = document.getElementById('searchinputcontainer');
```
saemideluxe, line 107: This is just for your information:

We can also use ```$(#searchinputcontainer)``` instead of ```document.getElementById('searchinputcontainer')```. We are using bliss.js, not jQuery, so performance is basically the same as using ```document.getElementById``` directly

bread/layout/components/forms/queryfield.py
saemideluxe: Okay, I saw that DjangQL enforces TEXTARE.

bread/layout/components/datatable.py
dward2nd: Ah... I didn't see this one. Thank you.

bread/layout/components/datatable.py
dward2nd: OK.

bread/layout/components/forms/queryfield.py
dward2nd: That's right.

bread/layout/components/forms/queryfield.py
dward2nd: OK. I didn't know we have that too because when I first saw `$`, I thought it would be jQuery so I avoid it.

bread/layout/components/forms/queryfield.py
dward2nd: This is because there are 2 input boxes that are switched back and forth. The normal search uses `input[type="text"]` while the advanced search uses `textarea`. There are also different icons for each mode and many different parts that has to directly modify within `Search`, so I decided to make a new one instead but use `Search` as the model component.

dward2nd:
> I think I want a few more changes. My biggest question would be: Is there a specific reason, that we cannot use a single input element to handle both cases?

In my opinion, I prefer it to be just a normal text box rather than `textarea`. But that would be interesting if we can just use `textarea` to handle searching in both ways. What do you think?

bread/layout/components/forms/queryfield.py
saemideluxe: Right, I would also recommend to avoid jQuery.
What we use is https://blissfuljs.com/ which I find much better as it tries to be compatible with vanilla javascript.

bread/querysetfield/__init__.py
dward2nd: I see. I changed this one so that it passes `**kwargs` to `super().contribute_to_class(cls, name, **kwargs`.

saemideluxe:
> > I think I want a few more changes. My biggest question would be: Is there a specific reason, that we cannot use a single input element to handle both cases?
> 
> In my opinion, I prefer it to be just a normal text box rather than `textarea`. But that would be interesting if we can just use `textarea` to handle searching in both ways. What do you think?

Yes, I think it would be a good idea to explore. I suggest you give it a try, but if it is more complicated code, then we can stay with two inputs.

bread/layout/components/forms/queryfield.py
saemideluxe: Okay, I see.

saemideluxe:
Closing this for now, might reopened when we get back to this feature.

PR #128, opened 2022-01-25
Feature/triggers

saemideluxe:
Initial working version, might need a few more features to be usable for concrete feature requests, but this is something we can start from.


bread/contrib/reports/urls.py
```diff
@@ -69,8 +69,8 @@
         Link(
             urls.reverse_model(Report, "browse"),
             label=_("Reports"),
-            iconname="download",
+            iconname="document",
         ),
-        menu.Group(_("Reports"), iconname="download"),
+        menu.Group(_("Reports"), iconname="document"),
     )
 )
```
saemideluxe, line 76: Commited to wrong branch, better icon for reports (I hope ;)

bread/urls.py
```diff
@@ -108,3 +109,7 @@
         browseview=administration.TaskResultBrowseView,
     ),
 ] + external_urlpatterns
+
+for app in apps.get_app_configs():
+    if app.name.startswith("bread.contrib"):
+        urlpatterns.append(path("contrib/", include(f"{app.name}.urls")))
```
saemideluxe, line 115: Saves us from registering all bread contrib urls manually (also wrong branch, should go to main)

bread/formatters.py
```diff
@@ -59,12 +58,6 @@ def as_text(value):
     return mark_safe(linebreaks(value, autoescape=True))
 
 
-def as_time(value):
-    if value.tzinfo:
-        value = value.astimezone(tz.gettz(settings.TIME_ZONE))
-    return f"{value.hour:02}:{value.minute:02}:{value.second:02}"
-
-
```
saemideluxe, line 67: Unused, has an unnecessary dependency to dateutil, Python stdlib and the dateparser library should do a better job.

bread/tests/settings.py
```diff
@@ -1,6 +1,10 @@
 from bread.settings.required import *  # noqa
 
-INSTALLED_APPS = BREAD_DEPENDENCIES + ["django.contrib.admin"]  # noqa
+INSTALLED_APPS = BREAD_DEPENDENCIES + [  # noqa
+    "django.contrib.admin",
+    "bread.contrib.reports",
+    "bread.contrib.triggers",
+]
```
saemideluxe, line 7: Should be here anyway, so the contribs are covered when running tests or doing migrations.

saemideluxe (commented):
Sorry for the unrelated changes, I had to switch branches every few hours during the last week and did a few wrong commits.

bread/contrib/triggers/urls.py
```diff
@@ -0,0 +1,46 @@
...
+
+from .models import DataChangeTrigger, DateFieldTrigger, SendEmail
+
+urlpatterns = []
+quickregister(
+    urlpatterns,
+    DataChangeTrigger,
+    editview=EditView._with(
+        fields=[
+            hg.H4(
```
wipascal, line 15: Is this a field? or is it just a trick for adding something arbitrary to the layout?

saemideluxe: Yes, a trick :D
Similar to datatable columns: When a simple string is passed it denotes a field which is derived from the model. If it is a "complex" object, then it will be used directly.

It makes prototyping easier as we do not have to override ```get_layout``` everytime.

bread/contrib/triggers/models.py
```diff
@@ -0,0 +1,170 @@
...
+    )
+    message = models.TextField(
+        _("Message"),
+        help_text=_(
+            "Will be rendered as a Django template with the name 'object' in the context"
+        ),
+    )
+
+    def run(self, object):
+        recepients = []
```
wipascal, line 67: typo: I think `recepients` should be `recipients`

bread/contrib/triggers/models.py
```diff
@@ -0,0 +1,170 @@
...
+    field = models.CharField(max_length=255)
+    offset_type = models.CharField(
+        max_length=255,
+        choices=tuple((name, value[1]) for name, value in INTERVAL_CHOICES.items()),
+    )
+    offset_amount = models.IntegerField(
+        help_text=_("Can be negative (before) or positive (after)")
+    )
+
+    def triggerdate(self, object):
```
wipascal, line 152: I would appreciate a typehint for what exactly this method returns

bread/contrib/triggers/models.py
saemideluxe: Thanks, fixed.

bread/contrib/triggers/apps.py
```diff
@@ -0,0 +1,60 @@
...
+def periodic_trigger():
+
+    from .models import DateFieldTrigger
+
+    for trigger in DateFieldTrigger.objects.filter(enable=True):
+        for instance in trigger.filter.queryset.all():
+            td = trigger.triggerdate(instance)
+            if (
+                td is not None
+                and timezone.now() <= td < timezone.now() + TRIGGER_PERIOD
```
wipascal, line 58: are we triggering the action before it's triggerdate is reached?

wipascal: what happens to the objects whose triggerdate is in the past which were not triggered for some reason?
If it is critical that each action is correctly triggered, I think this system is not very robust. We should have a way of knowing whether it was triggered on a certain instance or not, so that we can trigger the ones from the past which were missed... But maybe we should discuss this as soon as we have a concrete use case, because right now it is still quite an abstract idea

bread/contrib/triggers/models.py
saemideluxe: Added, although I was thinking the name would be clear enough.

bread/contrib/triggers/apps.py
saemideluxe: Right, if we have a trigger which fires on 2022-01-31 but our cronjob/celery task runs every hours we got 24 triggers. So we have to process "slices of time" and trigger things per slice.

bread/contrib/triggers/models.py
wipascal: the name `triggerdatetime` would have been clear enough :)

bread/contrib/triggers/apps.py
wipascal: Why are we not triggering into the past? I think it is un-intuitive for the user if a job is completed before it's time. If it's late, that's more understandeable

saemideluxe: I know, I have thought about this some.

The problem is that if you want to get it really right this will go down a very long rabbit hole. I think relying on the operating system (to keep the processes running) and rabbitmq/celery (for the scheduling and execution) should be enough for this kind of automated actions.

For a much more controlled way of handling certain workflows one could use ```bread.contrib.workflows```, as I have addresses a few of these kinds of problem there.

I will add the execution of actions as separate tasks though, that will at least create a trace for single failed actions.

saemideluxe: But need to test that first.

bread/contrib/triggers/models.py
saemideluxe: Well, I would prefer to *not* abuse names for type-hinting ;)

bread/contrib/triggers/apps.py
saemideluxe: Okay, can you take a look again?

wipascal: cool, thanks!

wipascal (approved):

PR #132, opened 2022-01-26
Use only 1 input element instead of two for BrowseView search

dward2nd:
As in #127, @saemideluxe suggested that there should only be one input element for the search box. So this pull request remove the existing `input[type='text']` for the normal search and use `textarea` for both modes instead.


dward2nd:
### Known issue

- Pre-filled text in search box displayed as an HTML escape string (like `&amp;`)

saemideluxe:
Okay, this feature requires way to much code for a feature which only advanced user will be able use.

Can you back leave this branch on the side for a moment and do some experiments back on main? (I copied your fix for escaping already to main)
- Try to add a second button, just before the close-button (should display correctly
- That button can be clicked to toggle the visibility of the djangoql helper-popup
- The search-bar should be created from the Search element, instead of creating a completely new component. Instead we can make the searchbar to support INPUT and TEXTAREA (in case of TEXTAREA just put "value" as a child instead of an attribute).
- The equal-sign at the begining can stay, auto-suggestion help from the djangoql widget seems to work even when we have it there.

saemideluxe:
Closing this for now, might reopend it when we get back to this feature.

PR #133, opened 2022-01-26
improve layout of PreferencesView

wipascal:


bread/views/globalpreferences.py
```diff
@@ -0,0 +1,70 @@
...
+from django.utils.module_loading import import_string
+from dynamic_preferences import views as preferences_views
+from dynamic_preferences.forms import GlobalPreferenceForm
+from dynamic_preferences.registries import global_preferences_registry
+
+from bread import layout as breadlayout
+from bread import menu
+
+
+class PreferencesView(SuccessMessageMixin, preferences_views.PreferenceFormView):
```
saemideluxe, line 15: Would it not make more sense to inherit from ```bread.views.util.BreadView``` and simply pass a form instance into the context (see https://django-dynamic-preferences.readthedocs.io/en/latest/quickstart.html#form-builder)? Possibly combined with ```bread.views.util.CustomFormMixin```, although would need to work around the expected ```model``` attribute.

wipascal: What is the advantage of using the form-builder instead of using the `GlobalPreferenceForm` as I do currently?

wipascal: I now extended `BreadView` to reduce code duplication. Would you like me to remove the subclassing of `preferences_views.PreferenceFormView`?

saemideluxe: Ah, I see. I was not sure if ```preferences_views.PreferenceFormView``` would work without any other method overrides. But if it works straight forward then keeping ```GlobalPreferenceForm``` and the ```preferences_views.PreferenceFormView``` is the better option.

I was somehow assuming that we have to overide get_form or get_form_class or similar. In such a case using the form-builder would have been more appropriate.

bread/views/globalpreferences.py
```diff
@@ -0,0 +1,42 @@
...
+                    hg.H3(section.capitalize()),
+                    *[
+                        breadlayout.forms.FormField(f)
+                        for f in section_fields.get(section)
+                    ],
+                )
+                for section in section_fields.keys()
+            ],
+            breadlayout.forms.helpers.Submit(),
+        )
```
saemideluxe, line 42: What do you think about using tabs here?

wipascal: I think it makes sense. It's done.

saemideluxe: Cool, thanks.

saemideluxe (approved):
Thanks, that is useful for a few other projects as well.

PR #138, opened 2022-02-07
add: better API to create links to instance-specific urls

saemideluxe:
The raw ModelHref is rather inconvenient to use, this should improved creating links to views with object-instances a bit.

### DIFF
--- a/bread/utils/links.py
+++ b/bread/utils/links.py
@@ -34,12 +34,12 @@ class LazyHref(hg.Lazy):
             kwargs["kwargs"] = {
                 k: hg.resolve_lazy(v, context) for k, v in kwargs["kwargs"].items()
             }
+        if "args" in kwargs:
+            kwargs["args"] = [hg.resolve_lazy(arg, context) for arg in kwargs["args"]]
         if "query" in kwargs:
             kwargs["query"] = {
                 k: hg.resolve_lazy(v, context) for k, v in kwargs["query"].items()
             }
-        if "args" in kwargs:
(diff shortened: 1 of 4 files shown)


bread/utils/links.py
```diff
@@ -58,30 +57,52 @@ class ModelHref(LazyHref):
...
+    def from_object(object: Union[models.Model, hg.Lazy], name: str, *args, **kwargs):
+        """
+        name: string which denotes an object action name as generated
+              by bread.utils.urls.model_urlname
+        object: instance of a Model, can be lazy
+        """
+
+        if "kwargs" not in kwargs:
+            kwargs["kwargs"] = {}
+        kwargs["kwargs"]["pk"] = object.pk
```
wipascal, line 96: Hmm `kwargs["kwargs"]`... Could we maybe rename one of these? Something like `kwargs["url_path_parameters"]`? (This would be a larger refactoring, I know. But the "kwargs" that are used for url path parameters was quite confusing for me when I first saw it.

wipascal:
very nice!

wipascal (approved):

bread/utils/links.py
saemideluxe: I would not really recommend it, although I know it is a bit confusing. The root of the problem is that Django has a function parameter called ```kwargs``` at https://docs.djangoproject.com/en/dev/ref/urlresolvers/#reverse. I would like to keep the name-compatatiliby with the original function, especially inside ```bread.utils.urls.reverse```. And the idea of the ```links``` module would be to abstract some of the details aways, at least for ```ModelHref```.

PR #140, opened 2022-02-08
add: flag to allow yearly triggers, help text and verbose names

saemideluxe:
Implemented, but needs testing.


saemideluxe:
Also, need to properly think about edge cases. I am not sure replacing the year with ```datetime.date.today().year``` is the correct solution.

PR #141, opened 2022-02-09
fix ObjectFieldValue would not work with fieldname containing "last"

wipascal:
if fieldname containing "last" it did not correctly call them method "last()"


wipascal:
hmm not sure it's an improvement. it's connected with https://github.com/basxsoftwareassociation/ponds_next/pull/39 but I am still thinking it through.

saemideluxe:
Can you give some more context? I think the issue revolves around the type of attribute or object, not explicitly about the term "last". "last" is also a call on django querysets, it could also be related to that.

PR #144, opened 2022-02-22
attempt to improve performance of pages with many modals

wipascal:
On my laptop it isn't very slow either way, so it would be interesting for me to know if it makes a difference on your machine


saemideluxe:
Great solution!

Old code

screenshot-1645546049_330x200


With change from PR

screenshot-1645546033_333x204

wipascal:
nice! :)

saemideluxe (approved):
Ok

PR #147, opened 2022-02-23
fix broken svgs

wipascal:


saemideluxe (approved):
Approved, but now you have to add your review as well :D

wipascal (commented):
great, thanks

PR #148, opened 2022-02-24
Feature/improve objectfieldvalue for onetomany

wipascal:
this is for the file downloads inside ponds tables


saemideluxe:
Okay, now after the fix on fix/svg I am not sure this is necessary anymore.

Also, as a side note, using ```str.join``` should be avoided in most places where we generate any HTML output. It basically enforces any ```BaseElement```s to be converted to strings, which should normally only be done by htmlgenerator's internal methods.

wipascal:
Yes, it's still necessary because it makes it possible to format "candidatereferences.file" as a list of downloadable files...

saemideluxe:
In that case I suggest we do not use ObjectFieldValue directly but just add a custom BaseElement-tree (or pass a custom formatter to ObjectFieldValue, but that might be less straight-forward).

The problem here is that ```ObjectFieldValue``` is a very generic element but ```join``` will not always work as expected, because it is not guaranteed that the resolved object is a string. And, maybe more importantly, when using ```mark_safe``` we should guarantee that the provided argument will never contain any html-strings, otherwise we have a potential XSS vulnerability.

We could theoretically also use ```hg.Iterator``` inside ```ObjectFieldValue.resolve``` to avoid those two issues. I am not too sure about that though. Sometimes we want a list of related items to be rendered as UL, sometimes with a comma, sometimes just with a single space, etc. To me it seems easier to just write something like 
```python
  hg.Iterator(
    hg.C("row.items"),
    "item",
    hg.BaseElement(hg.ObjectFieldValue("item", "file"), " ")
  )
```
or similar, instead of letting ObjectFieldValue do everything.

wipascal:
Yes, I agree. 

I think it will be better to just the `cell` parameter each time we want something as complex as `candidatereferences.file`, ok?

PR #149, opened 2022-02-25
make file download label parametrizeable

wipascal:
this is necessary because we don't want a label inside the ponds tables


saemideluxe (commented):
Is this used somewhere?

wipascal:
yes, in my ponds PR

saemideluxe (approved):

Issue #145, opened 2022-02-23
Add UI for user management


saemideluxe:
The django-admin backend has options to manage users, groups and permissions. We want to have something similar in our own frontend, inside the "Administration" menu.

- Add menu-item and view for "Users"
- Add menu-item and views for "Groups"

dward2nd:
Done #152

PR #152, opened 2022-04-08
User Management

dward2nd:
According to #145. I added three fundamental views, browse, read, and edit, for Django's `User` and `Group` models.


saemideluxe:
Closed as implemented in a simpler way already, closing.

